PR #1, opened 2026-04-07
Create skills for k8s API

alicefr:
Kubernetes community made available the API conventions at:
  https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md

Create skill for AI agent to help to review the ARCHITECTURE.md.

Assisted-by Claude

### DIFF
--- a/.agents/skills/k8s-api-review/SKILL.md
+++ b/.agents/skills/k8s-api-review/SKILL.md
@@ -1,0 +1,455 @@
+---
+name: k8s-api-review
+description: Review Kubernetes operators against official API conventions for CRD design, controller implementation, and architecture patterns
+---
+
+## What I do
+
+I perform a comprehensive design review of Kubernetes operators against the official [API Conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md). I check CRD definitions, controller implementations, and architecture patterns across 10 review areas: API structure, field design, validation, status conventions, references, behavioral conventions, advanced patterns, performance, security/RBAC, and observability. Each item is rated PASS, WARNING, or FAIL with actionable findings.
+
+## When to use me
+
+- Before submitting a new Kubernetes operator or CRD for review
+- When designing or iterating on custom resource APIs
+- To audit an existing operator for compliance with upstream conventions
+- After significant changes to CRD schemas or controller reconciliation logic
+- When preparing an operator for promotion from alpha to beta or GA
+
+---
+
+## Instructions for AI Agent
+
+Review the operator design and implementation against Kubernetes API conventions from https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md
+
+### How to Use This Skill
+
+**Step 1**: Identify the operator components to review:
+- Locate all CRD definitions (YAML files or Go types with `+kubebuilder` markers)
+- Locate controller/reconciler implementations
+- Review ARCHITECTURE.md or design documentation if available
+
+**Step 2**: For each CRD, apply the "CRD-Specific Review Template"
+
+**Step 3**: For each controller, apply the "Controller Review Template"
+
+**Step 4**: Apply all general checklist items (sections 1-10)
+
+**Step 5**: Generate the review report using the output format below
+
+### Review Process
+
+For each review area below, check the implementation against the criteria and report:
+- PASS: Compliant with convention
+- WARNING: Minor deviation or improvement opportunity
+- FAIL: Non-compliant, requires fix
+
+---
+
+## Review Checklist
+
+### 1. API Resource Structure
+
+#### 1.1 TypeMeta and ObjectMeta
+- [ ] All CRDs have `apiVersion` and `kind` fields
+- [ ] All CRDs have proper `metadata` section
+- [ ] Metadata includes `name`, `namespace` (if namespaced), `labels`, `annotations`
+- [ ] Uses `ownerReferences` for garbage collection (BootcNode -> BootcNodePool)
+
+#### 1.2 Spec and Status Separation
+- [ ] Clear separation between `spec` (desired state) and `status` (observed state)
+- [ ] Spec contains only user-provided configuration
+- [ ] Status contains only system-observed state
+- [ ] Status fields are never inputs to the system (read-only from user perspective)
+- [ ] No computed/derived fields in spec
+- [ ] Controllers write status, users write spec (distinct authorization scopes)
+
+#### 1.3 Resource Naming
+- [ ] Kind names are CamelCase and singular (e.g., `Database`, `Backup`, `Certificate`)
+- [ ] Resource names are lowercase and plural (e.g., `databases`, `backups`, `certificates`)
+- [ ] Group name follows domain format (e.g., `example.com`, `apps.mycompany.io`)
+- [ ] API version follows Kubernetes conventions (`v1alpha1`, `v1beta1`, `v1`)
+- [ ] Short names defined for common CLI usage (optional)
+
+### 2. Field Naming and Types
+
+#### 2.1 Field Naming Conventions
+- [ ] Field names use camelCase (not snake_case or kebab-case)
+- [ ] Boolean fields use positive names (e.g., `paused` not `notPaused`)
+- [ ] Time fields use suffix `Time` or include time units (e.g., `timeoutSeconds`)
+- [ ] Quantity fields include units in name (e.g., `maxUnavailable`, `timeoutSeconds`)
+- [ ] Reference fields use suffix `Ref` (e.g., `pullSecretRef`)
+
+#### 2.2 Primitive Types
+- [ ] Use `int32` or `int64` for integers (not `int`)
+- [ ] Use `string` for text
+- [ ] Use `metav1.Time` for timestamps (not `string`)
+- [ ] Use `resource.Quantity` for storage/memory quantities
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -2,3 +2,11 @@
 
 > [!WARNING]
 > Under construction. Come back later.
+
+## Documentation
+
+- **[Architecture](docs/ARCHITECTURE.md)** - System design and implementation details
+- **[Product Requirements](docs/PRD.md)** - Product vision and requirements
```
jlebon: I feel like this should be first in the list.

README.md
```diff
@@ -2,3 +2,11 @@
 
 > [!WARNING]
 > Under construction. Come back later.
+
+## Documentation
```
jlebon: This feels more like a TOC for agents. Perhaps it should live in `AGENTS.md` instead?

We'll eventually write a proper README here that's intended to be user-facing.

docs/K8S_API_CONVENTIONS.md
```diff
@@ -0,0 +1,553 @@
+# Kubernetes API Conventions - AI Agent Review Guide
```
jlebon: Of all the three docs added, this one is the crucial one. The others feel like too much overlap/repetition with this one. Can we start small and just land this one only for now? It's wordier than I would like, but fine too as is. (Basically I do wonder how much better writing all this out really is vs just pointing the agent at the source of truth--the API conventions doc URL--and telling it to ensure that the changed files conform to those conventions.)

Anyway, WDYT about structuring things like this:
- `AGENTS.md` has the TOC for agents
- New directory in e.g. `docs/ai/` (where this `K8S_API_CONVENTIONS.md` lives)
- Entries in TOC for each file in `docs/ai/` briefly describing the file and when to use it

docs/ARCHITECTURE.md
```diff
@@ -91,6 +91,42 @@ periodic heartbeats).
 
 ## CRDs
 
+### CRD Requirements
```
jlebon: Hmm, I wonder if we should just write out fully the `CustomResourceDefinition` object definitions themselves for each.

Ehhh, maybe not. I try to keep the architecture here to the point, and we elaborate on these kinds of things during the implementation plan and review at each checkpoint.

docs/ARCHITECTURE.md
```diff
@@ -91,6 +91,42 @@ periodic heartbeats).
...
+  subresources:
+    status: {}
+```
+
+This ensures:
+- Status updates don't increment `metadata.generation`
+- Spec and status can be updated independently via `/status` subresource
+- Proper RBAC separation (e.g., daemon can write status, not spec)
+
+### Field Types for CRD Schemas
```
jlebon: Is this section required? I think the added inline comments in the `BootcNodePool` and `BootcNode` examples should suffice.

docs/ARCHITECTURE.md
```diff
@@ -54,7 +54,7 @@ Two binaries: a **controller** (Deployment) and a **daemon** (DaemonSet).
 │  │                                                  │    │
 │  │  On spec change:                                 │    │
 │  │    if desiredImage != booted → bootc switch      │    │
-│  │    if staged && rebootApproved → reboot          │    │
+│  │    if staged && rebootApproval == Approved → reboot  │    │
```
jlebon: Broken ASCII art here.

docs/ARCHITECTURE.md
```diff
@@ -218,7 +218,7 @@ metadata:
     - kind: BootcNodePool  # owned by pool
 spec:                       # ← written by controller
   desiredImage: quay.io/example/myos@sha256:abc123
-  rebootApproved: false
+  rebootApproval: NotRequired  # NotRequired | Approved - set to Approved after drain, cleared after reboot
```
jlebon: I don't understand `NotRequired`.

docs/ARCHITECTURE.md
```diff
@@ -231,12 +231,12 @@ status:                     # ← written by daemon
     version: "9.4"
     timestamp: "2026-03-20T12:00:00Z"
     architecture: amd64
-    softRebootCapable: true
-    incompatible: false     # true if node has local mutations bootc can't manage
+    softRebootSupport: Supported  # Supported | Unsupported | Unknown
+    compatibility: Compatible     # Compatible | Incompatible | Unknown - Incompatible if local mutations exist
```
jlebon: But these are directly mapped from `bootc status --json`. I think it's cleaner to have the schema here be identical to what we get from bootc.

docs/ARCHITECTURE.md
```diff
@@ -146,7 +146,7 @@ spec:
     ref: quay.io/example/myos:latest
   rollout:
     maxUnavailable: 1        # IntOrString: int or percentage string (e.g. "25%")
-    paused: false            # when true, no new rollouts start
+    pauseMode: Active        # Active | Paused - when Paused, no new rollouts start
```
jlebon: Unsure about this one. `pauseMode: Active/Paused` feels awkward/potentially ambiguous vs `paused: false`.

How about:

```
state: Active | Paused
```

?

docs/ARCHITECTURE.md
jlebon: For this actually I think it's too imperative anyway.

I opened https://github.com/jlebon/bootc-operator/pull/3 to clean this up.

jlebon:
In discussions, you had also mentioned ensuring that the daemon can only update the status of its own CRD and that KubeVirt had figured out a way to do this. Were you able to investigate this further? Would be great if we could enforce that.

alicefr:
> In discussions, you had also mentioned ensuring that the daemon can only update the status of its own CRD and that KubeVirt had figured out a way to do this. Were you able to investigate this further? Would be great if we could enforce that.

Yes, this is the [CVE](https://github.com/kubevirt/kubevirt/security/advisories/GHSA-cp96-jpmq-xrr2). Basically, virt-handler (the privileged node component) if it was running on a compromised node could modify the node spec, and make the infrastracture pod being deployed eventually on the compromised node.

This isn't the case for us. However, I still think when we deal with upgrades we need to design the status in a way that a compromised daemon doesn't trigger the reboot or shutdown of other nodes. Either, we cannot trigger the reboot from the status or we can create a webhook that ensure that a daemon can only set the status on its dedicated node cr.

jlebon:
> This isn't the case for us. However, I still think when we deal with upgrades we need to design the status in a way that a compromised daemon doesn't trigger the reboot or shutdown of other nodes.

Moved this to https://github.com/jlebon/bootc-operator/issues/6 and had a suggestion there.

alicefr:
Create simple SKILL.md with actionable items to be used by an AI agent

.agents/skills/k8s-api-review/SKILL.md
jlebon, line 1: Hmm, this doesn't seem to be using the regular skill format? It's normally YAML frontmatter with a minimal description (this is always in-context) and then the doc (see e.g. https://opencode.ai/docs/skills/#use-an-example).

jlebon: Also, let's put this in e.g. `.agents/skills/<name>/SKILL.md`? (See https://opencode.ai/docs/skills/#place-files -- unfortunately, skill location is not standardized yet but at least OpenCode is trying to do the right thing there. I don't think Claude supports it yet, so feel free to add a Claude symlink to that...)

jlebon (commented):
Starting looking at this.

I also prompted Opus with:

> look at @api-conventions.md and then use it as well as the docs in kubebuilder repo to verify that every single check in @SKILL.md is justified by those documentation sources.

Do you agree that's a fair way to evaluate this? I want to double-check that all of those checks are grounded.

It's churning, so will let you know in the morning if it comes back with something interesting. :)

SKILL.md
```diff
@@ -0,0 +1,472 @@
...
+### 4. Status Conventions
+
+#### 4.1 Conditions
+- [ ] Status includes `conditions` array of type `[]metav1.Condition`
+- [ ] Each condition has: `type`, `status`, `reason`, `message`, `lastTransitionTime`, `observedGeneration`
+- [ ] Condition types are CamelCase (e.g., `UpToDate`, `Degraded`)
+- [ ] Condition reasons are CamelCase (e.g., `AllUpdated`, `RolloutInProgress`)
+- [ ] Conditions represent observations, not state machine transitions
+- [ ] Positive polarity conditions used (e.g., `Ready` not `NotReady`)
+- [ ] Long-running conditions (Ready, Available) come first in conditions array
```
jlebon: This is a weird one. Does that actually matter?

jlebon (commented):
Opus and myself didn't find anything alarmingly unreasonable there, so this looks sane! But let's fix the format and location.

SKILL.md
alicefr: You are right, it makes little sense

alicefr: Removed

.agents/skills/k8s-api-review/SKILL.md
jlebon: This still needs fixing.

alicefr:
I moved the skills under .agents/skills/k8s-api-review/SKILL.md, add a symlink for claude and changed the initial definition of the skills to match the [opencode example](https://opencode.ai/docs/skills/#use-an-example). It has the yaml header with the description and the sections "What I do" and "When to use me"

jlebon (approved):

PR #2, opened 2026-04-09
Create containerized cluster

alicefr:
Add VM-based Kubernetes cluster development environment with automated setup and management scripts.

  ## Usage                                                                                                                                        
  ```bash                                                                                                                                       
  # Build and start cluster                                                                                                                       
  make all cluster-start                                                                                                                          
   
  # Use kubectl from host                                                                                                                         
  export KUBECONFIG=./vm/kubeconfig                                                                                                             
  kubectl get nodes                                                                                                                               
                                                                                                                                                
  # Add worker nodes                                                                                                                              
  ./vm/join-node.sh node2                                                                                                                       
  ./vm/join-node.sh node3

### DIFF
--- a/containerfiles/images/Containerfile
+++ b/containerfiles/images/Containerfile
@@ -1,0 +1,6 @@
+FROM quay.io/fedora/fedora-bootc as root
+
+RUN dnf install -y kubernetes kubernetes1.35-kubeadm openssh-server qemu-guest-agent bind-utils iputils cloud-init && dnf clean all
+RUN passwd -d root
+
+RUN sed -i 's|"/opt/cni/bin"|"/var/lib/cni/bin"|g' /etc/crio/crio.conf
--- a/containerfiles/vm/entrypoint.sh
+++ b/containerfiles/vm/entrypoint.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+/usr/sbin/virtlogd &
+/usr/bin/virtstoraged &
+/usr/sbin/virtnetworkd -d
+/usr/sbin/virtqemud -v -t 0
--- a/docs/CLUSTER_SETUP.md
+++ b/docs/CLUSTER_SETUP.md
@@ -1,307 +1,95 @@
 # Kubernetes Cluster Setup Guide
 
-This guide explains how to create a Kubernetes cluster and add worker nodes using the VM-based development environment.
-
 ## Prerequisites
 
 - Podman installed
-- `bcvk` (bootc-virt-konverter) installed for converting bootc images to qcow2
-- Sufficient resources (8GB RAM per VM minimum)
+- `bcvk` (bootc-virt-konverter) for converting bootc images to qcow2
+- 8GB RAM per VM minimum
 
-## Building Images
-
-Before creating a cluster, build the required images and disk:
+## Quick Start
 
 ```bash
-# Build all images and create the bootc disk image
-make all
-```
-
-This will:
-1. Build the cluster container image (includes libvirt, qemu, virt-install)
-2. Build the fedora-bootc-k8s VM image (includes Kubernetes, CRI-O, cloud-init)
-3. Convert the bootc image to a qcow2 disk
-
-## Creating a New Cluster
-
-To create a fresh Kubernetes cluster with a single control plane node:
-
-```bash
-make cluster-start
-```
-
-This automated process will:
-1. Deploy the cluster container with `podman kube play`
-2. Setup the libvirt network with DNS (*.k8s.local domain)
-3. Create a VM named `node1` as the control plane
-4. Initialize Kubernetes on node1 with `kubeadm init`
-5. Install Calico CNI plugin for pod networking
-6. Configure kubectl for the core user
-7. Expose the API server on localhost:6443 (accessible from your host)
-8. Generate kubeconfig at `./vm/kubeconfig`
-
-The entire process takes 3-5 minutes. Once complete, you'll have a working single-node Kubernetes cluster accessible from both inside the VMs and from your host machine.
-
-### Verifying the Cluster
-
-Check that the cluster is running:
-
-```bash
-# SSH into the control plane node
-./vm/ssh-vm.sh node1
-
-# Inside the VM, check cluster status
-kubectl get nodes
-kubectl get pods -A
-```
-
-You should see:
-- node1 in `Ready` status
-- Calico pods running in `kube-system` namespace
-- Core components (etcd, kube-apiserver, kube-controller-manager, kube-scheduler) running
-
-## Accessing the Cluster from Your Host
-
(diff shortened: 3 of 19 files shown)


alicefr:
Still using privileged for networking, let try to switch to using passt instead of bridge mode

containerfiles/images/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/fedora/fedora-bootc as root
+
+RUN dnf install -y kubernetes kubernetes1.35-kubeadm openssh-server qemu-guest-agent bind-utils iputils cloud-init && dnf clean all
```
jlebon, line 3: A bit unsure about pulling in Fedora RPMs for this because I don't know how well-maintained it is. I had previously just used upstream bits: https://github.com/jlebon/bootc-operator/blob/ralph-2/tests/k8s/Containerfile.fcos.

But fine to give it a try!

Let's also pull in cri-o1.35.

containerfiles/images/Containerfile
```diff
@@ -0,0 +1,6 @@
+FROM quay.io/fedora/fedora-bootc as root
```
jlebon, line 1: I was thinking about starting from minimal here instead (would require a compose) to keep it as thin as possible.

We can look at that later.

containerfiles/images/Containerfile
jlebon: cloud-init 😢 
We can switch to Ignition once it's ready for fedora-bootc! :) (That's one of the main reasons why I was using FCOS for now but agree we should pivot to a more vanilla bootc image like fedora-bootc. Down the line, we could even add different distros there if we want.)

docs/CLUSTER_SETUP.md
```diff
@@ -0,0 +1,95 @@
+# Kubernetes Cluster Setup Guide
+
+## Prerequisites
+
+- Podman installed
+- `bcvk` (bootc-virt-konverter) for converting bootc images to qcow2
```
jlebon, line 6: Hum, that's not what bcvk stands for 🙃

jlebon: But on that note, I was thinking more of hiding this even more here and have it more automated. So basically, we'd have three container images total:
1. The harness container image with qemu/libvirt and logic for cluster bootstrapping/scaling/etc...
2. The bootable container image, based on fedora-bootc + kube bits (could imagine having multiple variants of this for kube versions or different bootc distros etc...).
3. The disk container image, containing just the qcow2 (it could be an OCI artifact but I really like the `podman build` semantics here to abstract the building step). The `Containerfile` for this would literally pull the bootable container image and run `bcvk install to-disk` and then a final scratch stage that just copies it in. I guess that's not unlike kubevirt's disk OCI images too.

And then when we run the harness, you can plug in the disk container image using an image mount.

But the idea then (and why I was thinking this should be a separate repo) is that all of these images would just be built e.g. daily and pushed to e.g. quay.io or ghcr.io and we just pull from there when developing and testing.

Obviously we make it easy to use your own custom local builds if e.g. you want to test with a custom bootc or whatever, but that's the idea. Because in the common case you're iterating on the operator so you're not necessarily changing the test harness itself that much. So let's do as much upfront work as we can so that CI testing for the operator spends the least amount of time on setting up the test environment. Simpler test instructions is also beneficial for coding agents of course.

containerfiles/vm/entrypoint.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
jlebon, line 1: So yeah, basically I was thinking of a much more powerful harness entrypoint here. Essentially fold in a lot of the functionality you have in `vm/` into the entrypoint. And ideally written in `!bash` :) Perhaps golang for consistency.

But essentially we'd have subcommands like (I was calling this `kvp` in my design docs):

```
kvp start    # starts cluster
kvp destroy  # destroys cluster
kvp add-node
kvp delete-node
...
```

That's the container image `ENTRYPOINT`, so you can do `podman run kvp start`. And then you'd do e.g. `podman exec $container kvp add-node` or whatever to access kvp functions after creation.

WDYT?

jlebon: (Actually, the `ENTRYPOINT` should definitely be some init process, e.g. `dumb-init -- kvp`.)

jlebon (commented):
High-level, I'd like to clearly separate all this test harness stuff from the bootc-operator. I was initially thinking making it a separate repo entirely, but perhaps we can start with them unified and then graduate to splitting them in the future if needed.

But anyway, for now let's put everything under... `tests/harness/`?

containerfiles/images/Containerfile
alicefr: Yeah I know, I don't think we can still use ignition with fedora, even if the work is in progress. But eventually we could reuse cloud-init with debian

containerfiles/vm/entrypoint.sh
alicefr: Yes. I agree we could develop a tool doing it. Bash scripts are a bit easier to debug to have the base infrastructure. Especially, the networking setup

alicefr:
> High-level, I'd like to clearly separate all this test harness stuff from the bootc-operator. I was initially thinking making it a separate repo entirely, but perhaps we can start with them unified and then graduate to splitting them in the future if needed.
> 
> But anyway, for now let's put everything under... `tests/harness/`?

If you want to put it under bvk we can. It didn't exist when I started or I missed it. But can we try to find an easier name as kvp :) it is hard to remember

docs/CLUSTER_SETUP.md
alicefr: > 3. The disk container image, containing just the qcow2 (it could be an OCI artifact but I really like the `podman build` semantics here to abstract the building step). The `Containerfile` for this would literally pull the bootable container image and run `bcvk install to-disk` and then a final scratch stage that just copies it in. I guess that's not unlike kubevirt's disk OCI images too.
> 

I tried to run bcvk in a containerfile and using podman remote for triggering the build on the host, but unfortunately bcvk expects to have access to the container storage locally

> And then when we run the harness, you can plug in the disk container image using an image mount.
> 

Yes, this is a nice improvement 

> But the idea then (and why I was thinking this should be a separate repo) is that all of these images would just be built e.g. daily and pushed to e.g. quay.io or ghcr.io and we just pull from there when developing and testing.
> 

+1
Do we want to create an organization or put this under bootc org?

containerfiles/images/Containerfile
alicefr: I see that ignition is available for fedora 43. I can try to switch to it

jlebon: No that's fine. Let's keep cloud-init for now. The Ignition work is ongoing in https://github.com/PeaceRebel/ignition/pulls. It'll eventually land in Ignition proper.

alicefr:
Create separate tool at https://github.com/alicefr/bink

PR #12, opened 2026-05-05
Add implementation plan and run kubebuilder init+create

jlebon:
See individual commits.

For the implementation plan, the one to focus on right now is milestone 1.

The second commit is pure kubebuilder generation. Though some files were removed in that same commit to keep them out of the object store; but otherwise everything else is left untouched for now. Another patch series will apply all the tweaks and fixes on top of the kubebuilder bits.


docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+- `DesiredImageState` enum type (`Staged`, `Booted`)
+
+### Validation (bink)
+
+- Start a bink cluster, `kubectl apply` the CRDs, create a BootcNodePool
+  and BootcNode, verify they're accepted and retrievable.
+
+## Milestone 2: Daemon
+
+Build the daemon first since the controller depends on daemon-reported
```
alicefr: Isn't the daemonset defined and deployed by the operator? You need to then add a manual temporary definition of the daemon

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+### 2d. DaemonSet manifest
+
+- `privileged: true`, `hostPID: true`
+- `nodeSelector: bootc.dev/managed: ""`
+- ServiceAccount + RBAC for BootcNode read/write
+
+### Validation (bink)
+
+- Start a bink cluster, deploy CRDs + DaemonSet
+- Manually create a BootcNode for a node, verify daemon populates status
```
alicefr: Instead of manually testing, we can create some ginkgo test. In this way, we can also review the validation step and we don't always need to to instruct the agent how to test

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+  `ImageInfo` struct, conditions)
+- Node condition type/reason constants (`Idle` with its reasons)
+- `DesiredImageState` enum type (`Staged`, `Booted`)
+
+### Validation (bink)
+
+- Start a bink cluster, `kubectl apply` the CRDs, create a BootcNodePool
+  and BootcNode, verify they're accepted and retrievable.
+
+## Milestone 2: Daemon
```
alicefr: I would split he milstone in 2 milestones. The first is the scaffolding of the daemon and that it sets the status correctly with its own validation steps, and then the state machine with the update

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+- Set `desiredImageState: Booted`, verify node reboots into new image
+  and daemon reports `Idle=True`
+
+## Milestone 3: Controller — Pool Reconciler
+
+### 3a. Pool membership sync
+
+- Watch BootcNodePool, Node (with predicates: label changes, Ready
+  condition, `spec.unschedulable` only), BootcNode (via ownerReference)
+- Match nodes via `nodeSelector`, create BootcNode with ownerReference
```
alicefr, line 118: wasn't the BootcNode owned by the Node instead of by the BootcNodePool?

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+- Start a bink cluster, deploy CRDs + DaemonSet
+- Manually create a BootcNode for a node, verify daemon populates status
+  from `bootc status`
+- Push an updated bootc image to the bink local registry
+- Set `desiredImage` to the new image, verify daemon stages it and
+  reports `Idle=False reason=Staged`
+- Set `desiredImageState: Booted`, verify node reboots into new image
+  and daemon reports `Idle=True`
+
+## Milestone 3: Controller — Pool Reconciler
```
alicefr, line 99: I have the impression that for milestone 3 we need a validation step at the end of each commit

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,206 @@
...
+- Update pool `spec.image.ref` to new digest
+- Verify rolling update: one node at a time cordoned, drained, rebooted,
+  uncordoned
+- Verify pool status counts and conditions update correctly throughout
+- Test pause/resume: pause mid-rollout, verify no new reboots, resume,
+  verify completion
+- Test conflict detection: create overlapping pool, verify
+  `Degraded/NodeConflict`
+
+## Milestone 4: Tag Resolution & Pull Secrets
```
alicefr: Here, I have the same feeling that each subsection requires its own validation step

alicefr:
nit: about commit 2, would be possible to commit everything was generated and then add a commit that deletes the files for sake of history. In this way, if I want to rerun those commands I can then apply the third commit to remove the unecessary files. I think it is also easier to review if we want to indeed keep certain files

alicefr:
Do we want to remove the tests folder? I think it something we want to implement and use for the validation step

jlebon:
Sorry, to reiterate here I didn't pay close attention to anything but milestone 1 for now. I think also I'd like to swap milestone 2 and 3 (i.e. work on the controller first). Perhaps I should actually _remove_ details from those milestones for now until we get there.

jlebon:
> Do we want to remove the tests folder? I think it something we want to implement and use for the validation step

Yeah, I'm actually adding it back in https://github.com/jlebon/bootc-operator/pull/13. When I initially reviewed the kubebuilder output, I was very aggressive in pruning things out. Guess I could fold this back into this commit to make the git history cleaner.

jlebon:
> nit: about commit 2, would be possible to commit everything was generated and then add a commit that deletes the files for sake of history. In this way, if I want to rerun those commands I can then apply the third commit to remove the unecessary files. I think it is also easier to review if we want to indeed keep certain files

Heh, I actually initially had it that way, but then squashed them together because of:

> Kubebuilder doesn't support selectively disabling features during
scaffolding so the following generated files were deleted as part
of this same commit to avoid carrying unnecessary content in the git
history:

Basically, as long as you have a commit with those files, they'll stay stored in git... forever. Which, I know it's not a lot of files/data, but still feels icky to do that early in the project. But this is also why I documented every single deleted file in the commit message, so that at least if you dod want to rerun the command, you could manually copy/paste that list into e.g. `xargs rm` to get the same effect.

docs/IMPLEMENTATION_PLAN.md
jlebon: > Isn't the daemonset defined and deployed by the operator?

Hmm, I don't think that's in the design. I know this is a common operator pattern though, and I could imagine adding a top-level operator which syncs the controller and daemonset manifests in the future.

docs/IMPLEMENTATION_PLAN.md
jlebon: s/ginkgo/envtest/ but yeah strong agree.

docs/IMPLEMENTATION_PLAN.md
jlebon: That's currently tracked at https://github.com/jlebon/bootc-operator/issues/4. I guess it's probably the right time to address that.

docs/IMPLEMENTATION_PLAN.md
jlebon: This is also making me realize that I need to add the actual test harness for bink as well so that it's not just "do these manual tests using bink", but "add these bink tests to the codebase" just like for envtests.

docs/IMPLEMENTATION_PLAN.md
jlebon: After [further reflection](https://github.com/jlebon/bootc-operator/issues/4#issuecomment-4398063585), I've decided to keep it as is.

docs/IMPLEMENTATION_PLAN.md
jlebon: Took that to https://github.com/jlebon/bootc-operator/issues/14. I'll leave it out for now and we can directly `kubectl create` from the Kustomize manifest.

jlebon:
OK, I've updated this now after sanity-checking and fixing up the whole implementation plan. Notably:
1. There is now an explicit milestone for the bink e2e test harness
2. We work on the controller before working on the daemon; in practice, a lot of it could be parallelized with enough hands
3. We validate things more progressively as things advance

alicefr:
@jlebon which images do we want to use for the tests? We should at least to have 2 version for testing upgrades. Not sure if this is out-of-scope to the plan or we should add a milestone after the first to introduce those images. Or do we want to include the test images to bink?

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,291 @@
...
+### 4d. fsnotify + polling
+
+- fsnotify watch on `/proc/1/root/ostree/bootc` for CHMOD events
+- Fallback: also try `/proc/1/root/sysroot/state/deploy` (for composefs
+  where `/ostree/bootc` doesn't exist)
+- Polling fallback every ~5 minutes
+- On event: re-read bootc status, update BootcNode.status if changed
+
+**Validation (e2e):** add a test that triggers an external bootc status
+change on the host (e.g. via SSH) and verifies the daemon detects it
```
alicefr, line 247: nit: if  we don't want to use ssh we can simply exec in the daemon pod and execute the command

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,291 @@
...
+`New(t, cfg)` handles the full lifecycle transparently. The `Config`
+struct maps to bink's cluster configuration (e.g. cluster name, VM
+memory) and can be extended as needed by later milestones.
+
+- Finds `bink` binary on PATH (overridable via `BINK_PATH` env var).
+  Fails the test if not found.
+- Runs `bink cluster start` with a unique cluster name per test
+  (e.g. `e2e-<short-hash>`, overridable via `Options`)
+- Waits for node1 to be Ready
+- Runs `bink api expose` to generate a kubeconfig
```
alicefr, line 80: nit: first you need to run `bink api expose` and then waiting for the node

docs/IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,291 @@
...
+accepted and retrievable. Validates both the harness and the CRDs on
+a real cluster.
+
+### Validation
+
+- Run the smoke test, verify it passes end-to-end.
+- Verify the cluster is torn down after test completes (including on
+  failure).
+
+## Milestone 3: Controller — Pool Reconciler
```
alicefr, line 99: I'm wondering if for doing the validation of this mileston we shouldn't simply using mocking instead of deploying a full cluster. It seems much cheaper to me and easier to implement and to test all the corner cases

jlebon: The majority of the validation in this milestone is indeed envtest-based exactly for this reason. There's just one check that enhances the e2e smoke test to sanity-check the whole thing in a real cluster.

docs/IMPLEMENTATION_PLAN.md
jlebon: Yeah, this is right in https://github.com/jlebon/bootc-operator/pull/16.

docs/IMPLEMENTATION_PLAN.md
jlebon: Good point!

jlebon:
> @jlebon which images do we want to use for the tests? We should at least to have 2 version for testing upgrades. Not sure if this is out-of-scope to the plan or we should add a milestone after the first to introduce those images. Or do we want to include the test images to bink?

For that I was thinking we do use the bink image, and for updates we just `FROM` the node image, tweak a file, and then commit with a new version label. We'll nail it down more once we get to those tests, but yeah I don't think bink itself needs to serve multiple versions just for testing updates.

PR #13, opened 2026-05-05
Fix up kubebuilder output

jlebon:
This builds on top of https://github.com/jlebon/bootc-operator/pull/12 to fix up the various kubebuilder bits before we go any further.

It's a lot of commits, but the majority are tiny and easy to review.


alicefr:
why not keeping ginkgo in [62aeca9](https://github.com/jlebon/bootc-operator/commit/62aeca97dc003a85c9cd548675449f5911e0d324) ?

alicefr:
Looks I would only keep ginkgo instead of replacing the with go test

jlebon:
> why not keeping ginkgo in [62aeca9](https://github.com/jlebon/bootc-operator/commit/62aeca97dc003a85c9cd548675449f5911e0d324) ?

I'm just not a fan of BDD-style tests and prefer simpler imperative tests. The diffstat on that commit is informative. We can still abstract things and keep it DRY with helpers and setup functions etc...

alicefr (approved):

PR #15, opened 2026-05-08
Fill in BootcNodePool and BootcNode CRDs

jlebon:
These are pretty straight-forward and map directly from the architecture doc.

Builds on top of #13


internal/controller/crd_test.go
```diff
@@ -0,0 +1,251 @@
...
+			Name: "workers",
+		},
+		Spec: bootcv1alpha1.BootcNodePoolSpec{
+			NodeSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"node-role.kubernetes.io/worker": "",
+				},
+			},
+			Image: bootcv1alpha1.ImageSpec{
+				Ref: "quay.io/example/myos:latest",
```
alicefr: nit: can we define a constant since it is used for validation later

internal/controller/crd_test.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	if got.Spec.Disruption.RebootPolicy != bootcv1alpha1.RebootPolicySoftReboot {
+		t.Errorf("rebootPolicy = %q, want %q", got.Spec.Disruption.RebootPolicy, bootcv1alpha1.RebootPolicySoftReboot)
+	}
+
+	// Update status
+	got.Status = bootcv1alpha1.BootcNodePoolStatus{
+		ObservedGeneration: got.Generation,
+		TargetDigest:       "sha256:abc123",
+		DeployedDigest:     "sha256:old789",
```
alicefr: same can we use constants

internal/controller/crd_test.go
```diff
@@ -0,0 +1,251 @@
...
+		t.Errorf("nodeCount = %d, want 3", got.Status.NodeCount)
+	}
+	if len(got.Status.Conditions) != 2 {
+		t.Errorf("len(conditions) = %d, want 2", len(got.Status.Conditions))
+	}
+
+	// Cleanup
+	if err := k8sClient.Delete(ctx, pool); err != nil {
+		t.Fatalf("Failed to delete BootcNodePool: %v", err)
+	}
```
alicefr: nit: this should be part of a defer. just after the creation If something fail it remains in the cluster

api/v1alpha1/bootcnode_types.go
```diff
@@ -20,38 +20,139 @@ import (
...
+
+	// softRebootCapable is true if this boot entry supports a
+	// userspace-only restart relative to the currently booted system.
+	// +optional
+	SoftRebootCapable bool `json:"softRebootCapable,omitempty"`
+
+	// incompatible is true if the boot entry has local mutations that
+	// bootc cannot manage (e.g. package layering via rpm-ostree).
+	// +optional
+	Incompatible bool `json:"incompatible,omitempty"`
```
alicefr, line 100: Does bootc return an error why this is incompatible? If so, maybe a string here is better then a bool

alicefr:
I haven't commented all the places in the tests but I would replace the strings with constants

alicefr:
We could use the Functional Options pattern in order to construct the object. If you start to define it here for the tests, it could serve as base for the next milestone.

Update: like [this example](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/informers/factory.go) from kubernetes

api/v1alpha1/bootcnode_types.go
jlebon: Nah, it's not an error to be incompatible. bootc can read the OSTree state and display it, but it can't e.g. upgrade.

jlebon:
Addressed comments and improved the CRD unit tests so it's more comprehensive.

PR #16, opened 2026-05-08
Add e2e test harness and CRD smoke test

jlebon:
This is where things get interesting CI-wise!

We hook up a test harness which uses bink. For now it's minimal but does
the fundamentals: it handles the cluster lifecycle and the test is just
handed a controller-runtime client.

Add a basic smoke test to exercise all this which just creates a
BootcNodePool and BootcNode and verifies they round-trip correctly.
That's not very different from the envtest version of this, but we'll
build on it in future milestones!

### DIFF
--- a/test/e2e/e2eutil/env.go
+++ b/test/e2e/e2eutil/env.go
@@ -1,0 +1,311 @@
+/*
+Copyright 2026.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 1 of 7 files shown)


jlebon:
Builds on top of #15

test/e2e/e2eutil/env.go
```diff
@@ -0,0 +1,295 @@
...
+	t.Helper()
+
+	binkPath := findBink(t)
+	clusterName := cfg.ClusterName
+	if clusterName == "" {
+		clusterName = generateClusterName(t)
+	}
+	memory := cfg.Memory
+	if memory == 0 {
+		memory = 4096
```
alicefr: nit: constant for default memory

test/e2e/e2eutil/env.go
```diff
@@ -0,0 +1,295 @@
...
+
+	// Kubeconfig is the path to the kubeconfig file for this cluster.
+	Kubeconfig string
+
+	// BinkPath is the resolved path to the bink binary.
+	BinkPath string
+
+	// Cluster is the bink cluster name.
+	Cluster string
+}
```
alicefr, line 67: If we decide to follow the functional optional pattern suggested by the comment: https://github.com/jlebon/bootc-operator/pull/15#issuecomment-4406644768 we could also apply it to the config and the env

PR #19, opened 2026-05-13
Various prep patches for controller tests

jlebon:
See individual commit messages!


test/e2e/e2eutil/env.go
```diff
@@ -97,6 +97,7 @@ func New(t *testing.T, opts ...Option) *Env {
 	kubeconfigPath := exposeAPI(t, clusterName)
 	applyManifests(t, kubeconfigPath, projectRoot)
 	k8sClient := buildClient(t, kubeconfigPath)
+	removeControlPlaneTaint(t, k8sClient, "node1")
```
alicefr: Instead of removing the taint from controlplanes, I think the controller should be allow to run on control plan. So, on the definition of the deployment we need to add the taint that make it schedulable on control planes as well.

alicefr:
I'm wondering if the cluster creation, the push of the image and the manifest application should sit outside the tests infrastructure. If we should move this part into the makefile. The main reason for is that in this way the tests remain generic and eventually they could be applied to an existing cluster (e.g. Openshift).

jlebon:
> I'm wondering if the cluster creation, the push of the image and the manifest application should sit outside the tests infrastructure. If we should move this part into the makefile. The main reason for is that in this way the tests remain generic and eventually they could be applied to an existing cluster (e.g. Openshift).

Heh I was thinking about that as well this morning. My main motivation though is more so that it can easily be deployed for manual developer testing.

Will whip something up.

jlebon:
> > I'm wondering if the cluster creation, the push of the image and the manifest application should sit outside the tests infrastructure. If we should move this part into the makefile. The main reason for is that in this way the tests remain generic and eventually they could be applied to an existing cluster (e.g. Openshift).
> 
> Heh I was thinking about that as well this morning. My main motivation though is more so that it can easily be deployed for manual developer testing.
> 
> Will whip something up.

Took this to https://github.com/jlebon/bootc-operator/issues/20.

PR #21, opened 2026-05-13
Add pool membership controller logic

jlebon:
This covers [milestone 3a](https://github.com/jlebon/bootc-operator/blob/main/docs/IMPLEMENTATION_PLAN.md#3a-pool-membership-sync) of the implementation plan.

See individual commits!


internal/controller/bootcnodepool_controller.go
```diff
@@ -53,13 +56,170 @@ type BootcNodePoolReconciler struct {
...
+}
+
+// removeBootcNode deletes a BootcNode for a node leaving the pool,
+// removes the managed label, and restores prior cordon state.
+func (r *BootcNodePoolReconciler) removeBootcNode(ctx context.Context, bn *bootcv1alpha1.BootcNode) error {
+	// Try to clean up the node (label + cordon state) before deleting
+	// the BootcNode. The node may have been deleted from the cluster.
+	var node corev1.Node
+	if err := r.Get(ctx, types.NamespacedName{Name: bn.Name}, &node); err == nil {
+		if err := r.restoreCordonState(ctx, &node); err != nil {
```
alicefr: I would add here a check if the node has a deletion timestamp, it would avoid to patch it since it is going to be removed. This might save a couple of API calls

internal/controller/bootcnodepool_controller.go
```diff
@@ -53,13 +56,170 @@ type BootcNodePoolReconciler struct {
...
+	}
+
+	newPullSecretRef := pool.Spec.PullSecretRef.DeepCopy()
+	if !reflect.DeepEqual(bn.Spec.PullSecretRef, newPullSecretRef) {
+		bn.Spec.PullSecretRef = newPullSecretRef
+		needUpdate = true
+	}
+
+	if needUpdate {
+		if err := r.Update(ctx, bn); err != nil {
```
alicefr: In general, we should always use patch instead of update because patch is able to detect the mismatches and eventually throw an error, Additionally, if the status has been updated by the daemon the update will probably error out

internal/controller/bootcnodepool_controller.go
```diff
@@ -156,6 +316,78 @@ func nodePredicates() predicate.Predicate {
...
+// bootc.dev/was-cordoned annotation. If the annotation is "true", the
+// node was already cordoned before the operator touched it, so we leave
+// it as is. Otherwise we uncordon it. The annotation is removed.
+func (r *BootcNodePoolReconciler) restoreCordonState(ctx context.Context, node *corev1.Node) error {
+	_, hasAnnotation := node.Annotations[bootcv1alpha1.AnnotationWasCordoned]
+	if !hasAnnotation {
+		return nil
+	}
+
+	patch := client.StrategicMergeFrom(node.DeepCopy())
```
alicefr: you are modifying the node object but what if the patch fails. You have updated the fields in the node object but it failed on server side. Doesn't this require a local copy for the modification?

internal/controller/membership_test.go
```diff
@@ -0,0 +1,299 @@
...
+	for {
+		if condFn() {
+			return
+		}
+		if time.Now().After(deadline) {
+			t.Fatalf("timed out waiting for: %s", msg)
+		}
+		time.Sleep(pollInterval)
+	}
+}
```
alicefr: Instead of re implementing the polling logic can we use [gomega](https://github.com/onsi/gomega)

internal/controller/membership_test.go
```diff
@@ -0,0 +1,299 @@
...
+			return k8sClient.Get(ctx, client.ObjectKey{Name: name}, &bn) == nil
+		}, "BootcNode "+name+" to be created")
+
+		// Check ownerReference.
+		owner := metav1.GetControllerOf(&bn)
+		if owner == nil {
+			t.Errorf("BootcNode %s has no controller owner", name)
+		} else if owner.Name != pool.Name {
+			t.Errorf("BootcNode %s owner = %s, want %s", name, owner.Name, pool.Name)
+		}
```
alicefr: Same here we could write a gomega match:
```golang
g.Expect(owner).NotTo(BeNil(), "BootcNode %s has no controller owner", name)
g.Expect(owner.Name).To(Equal(pool.Name), "BootcNode %s owner mismatch", name)
```

alicefr:
I'm really not a big fan on reimplementing the test logic for polling and the matches. It makes the code much harder to read. Also, with the increase of the test we will require the more and more complex logic.

alicefr:
I would also implement some unit testing for the reconcile loop for the bootcnode pool, there are a lot of corner cases there.

internal/controller/bootcnodepool_controller.go
```diff
@@ -36,28 +52,427 @@ type BootcNodePoolReconciler struct {
...
+
+	existing := apimeta.FindStatusCondition(pool.Status.Conditions, bootcv1alpha1.PoolDegraded)
+	if existing != nil && existing.Status == desired.Status && existing.Reason == desired.Reason && existing.Message == desired.Message {
+		// conflict condition status already matches desired
+		return nil
+	}
+
+	apimeta.SetStatusCondition(&pool.Status.Conditions, desired)
+	if err := r.Status().Update(ctx, pool); err != nil {
+		return fmt.Errorf("updating pool status: %w", err)
```
alicefr, line 254: We should use Patch instead of update. The status could be updated by the daemon as well. The update operation checks the `resourceVersion` of the entire object and it will error out

internal/controller/bootcnodepool_controller.go
```diff
@@ -36,28 +52,427 @@ type BootcNodePoolReconciler struct {
...
+			Type:   bootcv1alpha1.PoolDegraded,
+			Status: metav1.ConditionFalse,
+			Reason: bootcv1alpha1.PoolOK,
+		}
+	}
+
+	existing := apimeta.FindStatusCondition(pool.Status.Conditions, bootcv1alpha1.PoolDegraded)
+	if existing != nil && existing.Status == desired.Status && existing.Reason == desired.Reason && existing.Message == desired.Message {
+		// conflict condition status already matches desired
+		return nil
```
alicefr, line 249: nit: what about using DeepEqual for the existing and desired condition?

internal/controller/bootcnodepool_controller.go
```diff
@@ -36,28 +52,427 @@ type BootcNodePoolReconciler struct {
...
+	if err != nil {
+		return fmt.Errorf("listing matching nodes: %w", err)
+	}
+	matchingSet := map[string]*corev1.Node{}
+	for i := range matchingNodes {
+		matchingSet[matchingNodes[i].Name] = &matchingNodes[i]
+	}
+
+	// List all BootcNodes and partition into owned by this pool vs others.
+	allBootcNodes, err := r.listAllBootcNodes(ctx)
```
alicefr, line 98: Instead of listing all the bootc node we could filter the bootcnode that already belong to the pool by adding a label to the bn at creation. It saves some memory of getting only the node owned by that pool.

internal/controller/bootcnodepool_controller.go
jlebon: Yes agree. I do use `Patch` elsewhere. It wasn't clear to me whether `Update` was safe here because the status is a subresource but it seems like they do still share the same `resourceVersion`.

Anyway, fixed!

internal/controller/bootcnodepool_controller.go
jlebon: In all the cases we currently `Patch()`, we bubble up the error all the way up to `Reconcile()` so we're never really reading the outdated objects. That said, this is indeed a bug waiting to happen, so I've switched over all `Patch` instances to handle this correctly.

internal/controller/bootcnodepool_controller.go
jlebon: We need all of them for conflict detection.

(Marking as resolved for now. Feel free to unresolve to continue the conversation!)

internal/controller/bootcnodepool_controller.go
jlebon: That whole logic is cleaned up in a follow-up patch series, so leaving it as is for now. (And we do indeed use `DeepEqual` there.)

(You can now see this in https://github.com/jlebon/bootc-operator/pull/22.)

jlebon:
Thanks for the review! Updated for comments. I tried to fold all the changes into their respective commits but kept the gomega switchover as a separate commit to avoid that pain.

> I'm really not a big fan on reimplementing the test logic for polling and the matches. It makes the code much harder to read. Also, with the increase of the test we will require the more and more complex logic.

Thanks, I wasn't deeply familiar with gomega and you made me take a closer look. I was under the impression that it was a part of ginkgo, but no it clearly can work without it. Switched over to that now!

> I would also implement some unit testing for the reconcile loop for the bootcnode pool, there are a lot of corner cases there.

Hmm, this patch series does add quite a few envtests here though. Is there anything obvious that's missing?

internal/controller/crd_test.go
```diff
@@ -145,68 +131,69 @@ func TestBootcNodeCRD(t *testing.T) {
...
 
 	node := testutil.NewNode("invalid-image-state", testImageDigestRefA)
 	node.Spec.DesiredImageState = "Invalid"
-	if err := k8sClient.Create(ctx, node); err == nil {
+	err := k8sClient.Create(ctx, node)
+	if err == nil {
 		_ = k8sClient.Delete(ctx, node)
-		t.Fatal("Expected creation with invalid desiredImageState to fail, but it succeeded")
 	}
+	g.Expect(err).To(HaveOccurred(), "Expected creation with invalid desiredImageState to fail")
```
alicefr: nit: can we assert the error we expect

internal/controller/crd_test.go
```diff
@@ -145,68 +131,69 @@ func TestBootcNodeCRD(t *testing.T) {
...
 		testutil.WithWorkerSelector(),
 		testutil.WithRebootPolicy("Invalid"),
 	)
-	if err := k8sClient.Create(ctx, pool); err == nil {
+	err := k8sClient.Create(ctx, pool)
+	if err == nil {
 		_ = k8sClient.Delete(ctx, pool)
-		t.Fatal("Expected creation with invalid rebootPolicy to fail, but it succeeded")
 	}
+	g.Expect(err).To(HaveOccurred(), "Expected creation with invalid rebootPolicy to fail")
```
alicefr: nit: can we check the error we expect

internal/controller/crd_test.go
```diff
@@ -145,68 +131,69 @@ func TestBootcNodeCRD(t *testing.T) {
...
 	ctx := context.Background()
 
 	pool := testutil.NewPool("empty-image-ref", "", testutil.WithWorkerSelector())
-	if err := k8sClient.Create(ctx, pool); err == nil {
+	err := k8sClient.Create(ctx, pool)
+	if err == nil {
 		_ = k8sClient.Delete(ctx, pool)
-		t.Fatal("Expected creation with empty image.ref to fail, but it succeeded")
 	}
+	g.Expect(err).To(HaveOccurred(), "Expected creation with empty image.ref to fail")
```
alicefr: nit: same let's check a specific error

internal/controller/membership_test.go
```diff
@@ -90,130 +71,112 @@ func TestMembershipCreatesBootcNodes(t *testing.T) {
...
-		t.Fatalf("Failed to patch node labels: %v", err)
-	}
+	g.Expect(k8sClient.Patch(ctx, &fresh, patch)).To(Succeed())
 
 	// Wait for BootcNode to be deleted.
-	waitForDeleted(t, "mem-worker-1", &bootcv1alpha1.BootcNode{})
+	g.Eventually(func() bool {
+		err := k8sClient.Get(ctx, client.ObjectKey{Name: "mem-worker-1"}, &bootcv1alpha1.BootcNode{})
+		return apierrors.IsNotFound(err)
+	}).Should(BeTrue())
```
alicefr: just return the error and evaluate the error is IsNotFound. In this way, the error type is included in the error message if it doesn't match

test/e2e/controller_test.go
```diff
@@ -41,53 +41,47 @@ const (
...
-			return false, fmt.Errorf("getting node1: %w", err)
-		}
-		_, ok := node.Labels[bootcv1alpha1.LabelManaged]
-		return ok, nil
-	})
+	g.Eventually(func(g Gomega) {
+		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: "node1"}, &node)).To(Succeed())
+		g.Expect(node.Labels).To(HaveKey(bootcv1alpha1.LabelManaged))
+	}).Should(Succeed())
 }
```
alicefr:
```suggestion
g.Eventually(func() (map[string]string, error) {
    err := env.Client.Get(ctx, client.ObjectKey{Name: "node1"}, &node)
    return node.Labels, err
}).Should(HaveKey(bootcv1alpha1.LabelManaged))

File: test/e2e/e2eutil/env.go [hunk 14]
```diff
@@ -397,23 +393,25 @@ func waitForNodeReady(t *testing.T, c client.Client, nodeName string) {
...
 			if cond.Type == corev1.NodeReady && cond.Status == corev1.ConditionTrue {
-				t.Logf("  node %q is Ready", nodeName)
-				return true, nil
+				ready = true
+				break
 			}
 		}
-		return false, nil
-	})
+		g.Expect(ready).To(BeTrue(), "node %q not Ready yet", nodeName)
```
alicefr:
```suggestion
g.Expect(node.Status.Conditions).To(ContainElement(And(
    HaveField("Type", corev1.NodeReady),
    HaveField("Status", corev1.ConditionTrue),
)), "node %q not Ready yet", nodeName)

File: internal/controller/membership_test.go [hunk 15]
```diff
@@ -246,4 +246,30 @@ func TestMembershipConflictDetection(t *testing.T) {
...
+
+	// Verify pool2 recovers: Degraded condition should clear.
+	g.Eventually(func(g Gomega) {
+		var p bootcv1alpha1.BootcNodePool
+		g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: pool2.Name}, &p)).To(Succeed())
+		cond := apimeta.FindStatusCondition(p.Status.Conditions, bootcv1alpha1.PoolDegraded)
+		g.Expect(cond).NotTo(BeNil())
+		g.Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+		g.Expect(cond.Reason).To(Equal(bootcv1alpha1.PoolOK))
+	}).Should(Succeed())
```
alicefr:
```suggestion
g.Eventually(func() ([]metav1.Condition, error) {
    var p bootcv1alpha1.BootcNodePool
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(pool2), &p)
    return p.Status.Conditions, err
}).Should(ContainElement(And(
    HaveField("Type", bootcv1alpha1.PoolDegraded),
    HaveField("Status", metav1.ConditionFalse),
    HaveField("Reason", bootcv1alpha1.PoolOK),
)))

File: internal/controller/membership_test.go [hunk 16]
```diff
@@ -246,4 +246,30 @@ func TestMembershipConflictDetection(t *testing.T) {
...
+
+	// Verify node3 is now owned by pool2.
+	g.Eventually(func(g Gomega) {
+		var bn bootcv1alpha1.BootcNode
+		g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: node3.Name}, &bn)).To(Succeed())
+		owner := metav1.GetControllerOf(&bn)
+		g.Expect(owner).NotTo(BeNil())
+		g.Expect(owner.Name).To(Equal(pool2.Name))
+	}).Should(Succeed())
```
alicefr:
```suggestion
g.Eventually(func() (*metav1.OwnerReference, error) {
    var bn bootcv1alpha1.BootcNode
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(node3), &bn)
    return metav1.GetControllerOf(&bn), err
}).Should(And(
    Not(BeNil()),
    HaveField("Name", pool2.Name),
))

alicefr:
> Hmm, this patch series does add quite a few envtests here though. Is there anything obvious that's missing?

We are not testing the path when a node is deleted and that the corresponding bootc should be deleted as well.

internal/controller/membership_test.go
jlebon: OK so gomega has some magic convention here when the `Eventually` closure returns two parameters. Interesting...

jlebon:
Updated for comments!

> > Hmm, this patch series does add quite a few envtests here though. Is there anything obvious that's missing?
> 
> We are not testing the path when a node is deleted and that the corresponding bootc should be deleted as well.

Added!

alicefr (approved):

PR #22, opened 2026-05-20
Lay some groundworks for rollout logic

jlebon:
I.e. this is the very start of milestone 3b, though a lot of it is cleaning things up and clarifying the architecture.

See individual commits.

This builds on top of https://github.com/jlebon/bootc-operator/pull/21.


internal/controller/bootcnodepool_controller.go
```diff
@@ -73,12 +74,30 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+// setInvalidSpecCondition sets Degraded/InvalidSpec on the pool and
+// returns (Result, nil) so Reconcile stops without requeueing.
+func (r *BootcNodePoolReconciler) setInvalidSpecCondition(ctx context.Context, pool *bootcv1alpha1.BootcNodePool, specErr error) (ctrl.Result, error) {
+	apimeta.SetStatusCondition(&pool.Status.Conditions, metav1.Condition{
+		Type:    bootcv1alpha1.PoolDegraded,
+		Status:  metav1.ConditionTrue,
+		Reason:  bootcv1alpha1.PoolInvalidSpec,
+		Message: specErr.Error(),
+	})
+	if err := r.Status().Update(ctx, pool); err != nil {
```
alicefr, line 334: I would use patch here

internal/controller/bootcnodepool_controller.go
```diff
@@ -73,12 +74,30 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 }
 
+// setInvalidSpecCondition sets Degraded/InvalidSpec on the pool and
+// returns (Result, nil) so Reconcile stops without requeueing.
+func (r *BootcNodePoolReconciler) setInvalidSpecCondition(ctx context.Context, pool *bootcv1alpha1.BootcNodePool, specErr error) (ctrl.Result, error) {
+	apimeta.SetStatusCondition(&pool.Status.Conditions, metav1.Condition{
+		Type:    bootcv1alpha1.PoolDegraded,
+		Status:  metav1.ConditionTrue,
+		Reason:  bootcv1alpha1.PoolInvalidSpec,
+		Message: specErr.Error(),
```
alicefr, line 332: Don't we also need to set the LastTransitionTime?

api/v1alpha1/bootcnode_types.go
```diff
@@ -43,9 +47,16 @@ const (
...
+// NodeDegraded condition reasons.
+const (
+	// NodeReasonError is a catch-all reason for daemon errors. The
+	// condition message provides details about the specific error.
+	NodeReasonError string = "Error"
 
-	// NodeReasonStagingFailed means something went wrong during staging.
-	NodeReasonStagingFailed string = "StagingFailed"
+	// NodeReasonOK means no errors.
+	NodeReasonOK string = "OK"
```
alicefr: can we find something more expressive like Healthy or Ready?

internal/controller/bootcnodepool_controller.go
jlebon: This is handled by `SetStatusCondition`.

internal/controller/bootcnodepool_controller.go
jlebon: Hmm, is that really necessary? By definition, the controller owns the status subresource.

jlebon:
Rebased and renamed OK --> Healthy!

internal/controller/bootcnodepool_controller.go
alicefr: But the status is updated by the daemon as well, so there might be conflicts with the resource version if daemon has updated the status

jlebon: The daemon only updates the BootcNode status.

alicefr: ah right, I have somehow mixed with the node, good point!

api/v1alpha1/bootcnodepool_types.go
```diff
@@ -98,6 +98,12 @@ type RolloutSpec struct {
 	// continues and status.targetDigest is kept current.
 	// +optional
 	Paused bool `json:"paused,omitempty"`
+
+	// drainTimeout is the maximum duration to wait for a node drain to
+	// complete before marking the node degraded and freeing the reboot
+	// slot. If not set, drains block indefinitely.
+	// +optional
+	DrainTimeout *metav1.Duration `json:"drainTimeout,omitempty"`
```
alicefr: You are missing the unit of measurment in the name

internal/controller/bootcnodepool_controller.go
```diff
@@ -492,3 +515,17 @@ func nodeReadyStatus(node *corev1.Node) corev1.ConditionStatus {
 func nodeUnschedulableChanged(oldNode, newNode *corev1.Node) bool {
 	return oldNode.Spec.Unschedulable != newNode.Spec.Unschedulable
 }
+
+// invalidSpecError indicates a user-provided spec value that the
+// controller cannot process. Reconcile surfaces these as
+// Degraded/InvalidSpec conditions rather than requeueing with backoff.
+type invalidSpecError struct {
```
alicefr: nit: we could add a constructor for the new error type

internal/controller/membership_test.go
```diff
@@ -39,6 +39,50 @@ const (
...
+
+	// Wait for pool to be marked Degraded/InvalidSpec.
+	g.Eventually(func(g Gomega) {
+		var p bootcv1alpha1.BootcNodePool
+		g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: pool.Name}, &p)).To(Succeed())
+		cond := apimeta.FindStatusCondition(p.Status.Conditions, bootcv1alpha1.PoolDegraded)
+		g.Expect(cond).NotTo(BeNil())
+		g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
+		g.Expect(cond.Reason).To(Equal(bootcv1alpha1.PoolInvalidSpec))
+	}).Should(Succeed())
```
alicefr:
```suggestion
g.Eventually(func() ([]metav1.Condition, error) {
    var p bootcv1alpha1.BootcNodePool
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(pool), &p)
    return p.Status.Conditions, err
}).Should(ContainElement(And(
    HaveField("Type", bootcv1alpha1.PoolDegraded),
    HaveField("Status", metav1.ConditionTrue),
    HaveField("Reason", bootcv1alpha1.PoolInvalidSpec),
)))

File: internal/controller/membership_test.go [hunk 7]
```diff
@@ -39,6 +39,50 @@ const (
...
+
+	// Wait for pool to recover: Degraded=False/OK.
+	g.Eventually(func(g Gomega) {
+		var p bootcv1alpha1.BootcNodePool
+		g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: pool.Name}, &p)).To(Succeed())
+		cond := apimeta.FindStatusCondition(p.Status.Conditions, bootcv1alpha1.PoolDegraded)
+		g.Expect(cond).NotTo(BeNil())
+		g.Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+		g.Expect(cond.Reason).To(Equal(bootcv1alpha1.PoolOK))
+	}).Should(Succeed())
```
alicefr:
```suggestion
g.Eventually(func() ([]metav1.Condition, error) {
    var p bootcv1alpha1.BootcNodePool
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(pool), &p)
    return p.Status.Conditions, err
}).Should(ContainElement(And(
    HaveField("Type", bootcv1alpha1.PoolDegraded),
    HaveField("Status", metav1.ConditionFalse),
    HaveField("Reason", bootcv1alpha1.PoolOK),
)))

File: internal/controller/bootcnodepool_controller.go [hunk 8]
```diff
@@ -608,3 +613,123 @@ func syncConflictCondition(pool *bootcv1alpha1.BootcNodePool, conflictingPools [
...
+	}
+}
+
+// classifyNode determines the effective state of a BootcNode.
+func classifyNode(bn *bootcv1alpha1.BootcNode) (nodeState, error) {
+	// Check Degraded first — takes priority over activity state.
+	degradedCond := apimeta.FindStatusCondition(bn.Status.Conditions, bootcv1alpha1.NodeDegraded)
+	if degradedCond != nil && degradedCond.Status == metav1.ConditionTrue {
+		return nodeStateDegraded, nil
+	}
```
alicefr:
```suggestion
if apimeta.IsStatusConditionTrue(bn.Status.Conditions, bootcv1alpha1.NodeDegraded) {
    return nodeStateDegraded, nil
}

File: internal/controller/bootcnodepool_controller.go [hunk 9]
```diff
@@ -608,3 +613,123 @@ func syncConflictCondition(pool *bootcv1alpha1.BootcNodePool, conflictingPools [
...
+
+	if bn.Status.Booted == nil {
+		return 0, fmt.Errorf("no booted deployment")
```
alicefr: not idle? Or do we expect the Booted to be immediately set by the daemon?

internal/controller/bootcnodepool_controller.go
```diff
@@ -608,3 +613,123 @@ func syncConflictCondition(pool *bootcv1alpha1.BootcNodePool, conflictingPools [
...
+		// Image matches; nothing for the controller to act on
+		// regardless of whether the daemon has settled yet.
+		return nodeStateIdle, nil
+	}
+
+	// OK, the node isn't yet booting the desired digest. Let's dig into where
+	// it is in the update cycle.
+
+	idleCond := apimeta.FindStatusCondition(bn.Status.Conditions, bootcv1alpha1.NodeIdle)
+	if idleCond != nil && idleCond.Status == metav1.ConditionFalse {
```
alicefr:
```suggestion
if apimeta.IsStatusConditionFalse(bn.Status.Conditions, bootcv1alpha1.NodeIdle) {

File: internal/controller/bootcnodepool_controller_test.go [hunk 11]
```diff
@@ -21,8 +21,129 @@ import (
...
+			Reason: reason,
+		}
+	}
+
+	tests := []struct {
+		name         string
+		bootedDigest string
+		conditions   []metav1.Condition
+		want         nodeState
+		wantErr      bool
```
alicefr: nit: what if instead of checking if there is an error, we check the exact error
```suggestion
wantedErr *error

File: internal/controller/bootcnodepool_controller_test.go [hunk 12]
```diff
@@ -21,8 +21,129 @@ import (
...
+			}
+			bn.Name = nodeName
+			if tt.bootedDigest != "" {
+				bn.Status.Booted = &bootcv1alpha1.ImageInfo{
+					ImageDigest: tt.bootedDigest,
+				}
+			}
+			got, err := classifyNode(bn)
+			if tt.wantErr {
+				g.Expect(err).To(HaveOccurred())
```
alicefr: nit: if we change the error to the error type, this becomes:
```golang
g.Expect(err).To(Equal(tt.wantedErr))
```

internal/controller/membership_test.go
alicefr:
```suggestion
g.Eventually(func() ([]metav1.Condition, error) {
    var p bootcv1alpha1.BootcNodePool
    err := k8sClient.Get(ctx, client.ObjectKeyFromObject(pool), &p)
    return p.Status.Conditions, err
}).Should(ContainElement(And(
    HaveField("Type", bootcv1alpha1.PoolDegraded),
    HaveField("Status", metav1.ConditionTrue),
    HaveField("Reason", bootcv1alpha1.PoolInvalidSpec),
)))

File: internal/controller/bootcnodepool_controller.go [hunk 13]
```diff
@@ -332,10 +357,15 @@ func (r *BootcNodePoolReconciler) syncBootcNodeSpec(ctx context.Context, pool *b
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *BootcNodePoolReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.Recorder = mgr.GetEventRecorder("bootc-operator")
```
alicefr: You don't generate any events do you want to add this here?

config/rbac/role.yaml
```diff
@@ -4,6 +4,13 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - create
+  - patch
```
alicefr, line 13: nit: do we want to add the events in this PR or when they are actually created

alicefr:
@jlebon the PR looks overall good. My comments are mostly nits or small questions. When they are addressed feel free to merge it

api/v1alpha1/bootcnodepool_types.go
jlebon: Good point. `metav1.Duration` can deserialize strings like "30s" and "1h30m" so there's no single unit really. Kured does it this way as well. But the huge disadvantage of that is that the controller then has to do this deserialization and handle errors instead of letting the API server do it at admission time (and I don't really want to write a validate regex for it).

Anyway, let's just match core K8s APIs and switch to seconds.

internal/controller/bootcnodepool_controller.go
jlebon: Good point. I think idle is probably the cleanest here. It shouldn't stop us from progressing on other nodes, so an error feels incorrect indeed. I added a comment as well that ideally in the future we detect stuck nodes like this. (Similar to the comment below.)

internal/controller/bootcnodepool_controller_test.go
jlebon: The switch to idle actually allows us to drop the need for `wantErr` entirely!

internal/controller/bootcnodepool_controller.go
jlebon: I agree in spirit, though we're churning so quickly here, the code that uses it is not far behind (already starting to get the ball rolling on that one locally... heh).

Will keep it for now to avoid the conflict-fest of extricating it.

jlebon:
Bypassing review requirement based on implicit approval in https://github.com/jlebon/bootc-operator/pull/22#issuecomment-4505908471.

PR #23, opened 2026-05-20
tree: add Apache 2.0 license, switch to SPDX headers

jlebon:
The tree was missing a LICENSE file, but all the code was effectively
already Apache 2.0 as declared in their headers. While we're here,
replace those with SPDX instead.


jlebon:
(Builds on top of https://github.com/jlebon/bootc-operator/pull/22 which adds a new file but could break it out too.)

alicefr (approved):

PR #25, opened 2026-05-21
Scaffolding for the bootc operator daemon

alicefr:
This PR introduces the daemon for the bootc operator. This is simply the framework for implementing the logic at a later stage.


alicefr:
I want to test it with the deploy. I have only verified with the e2e tests.

cmd/daemon/main.go
```diff
@@ -0,0 +1,61 @@
+/*
```
jlebon: Let's rebase this on top of https://github.com/jlebon/bootc-operator/pull/23 and use SDPX-License-Identifier here.

jlebon: Also, let's move `cmd/main.go` to `cmd/controller/main.go`.

cmd/daemon/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		setupLog.Error(err, "Failed to start manager")
+		os.Exit(1)
+	}
+
+	setupLog.Info("starting daemon")
```
jlebon: Starting

for consistency

config/daemon/daemon.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: daemon
+  namespace: system
+  labels:
+    app.kubernetes.io/name: bootc-operator
+    app.kubernetes.io/component: daemon
```
jlebon, line 8: Hmm, this label doesn't exist on the controller. Either we add it there or remove this one here for consistency.

Similarly below.

config/rbac/daemon_service_account.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
```
jlebon, line 1: Perhaps let's rename the controller's SA file to `controller_service_account.yaml` for consistency.

Containerfile.daemon
```diff
@@ -0,0 +1,19 @@
+FROM quay.io/fedora/fedora-minimal:44 AS buildroot
```
jlebon: I think I'd rather match the MCO and just shove all the binaries in the same image. And then the manifests encode the right entrypoint.

That should simplify a bunch of things.

jlebon (commented):
Let's try to get https://github.com/jlebon/bootc-operator/pull/29 in first and rebase on top of this. Combined with changing to using a single image as mentioned, this should make this change simpler.

alicefr:
@jlebon before merging this, I think we need to decide which option we want for https://github.com/jlebon/bootc-operator/issues/14. Since, this introduce the daemon infrastructure, it will be much easier to review it here

cmd/daemon/main.go
```diff
@@ -0,0 +1,47 @@
...
+
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		setupLog.Error(err, "Failed to start manager")
+		os.Exit(1)
+	}
+
+	setupLog.Info("Starting")
```
jlebon: Ahh sorry, I just meant capitalizing the S. "Starting daemon" is still fine.

jlebon (approved):
LGTM! Thanks. Just pushed a minor tweak and rebased.

I'll rebase https://github.com/jlebon/bootc-operator/pull/29 on top of this.

PR #31, opened 2026-05-26
ci: fetch versioned bink release instead of building from source

### LINKED ISSUE #30: Use versioned bink
I just created the first version of bink [v0.1.0](https://github.com/alicefr/bink/releases/tag/v0.1.0). Move the CI from using main to the version 0.1.0 and remove the bink compilation

alicefr:
Fetch bink instead of compiling.
Fixes: #30

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [main]
 
+env:
+  BINK_VERSION: v0.1.1
+
 concurrency:
   group: ci-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [main]
 
+env:
+  BINK_VERSION: v0.1.0
```
jlebon: Filed https://github.com/jlebon/bootc-operator/issues/32 related to this.

.github/workflows/ci.yaml
```diff
@@ -32,33 +35,14 @@ jobs:
       - name: Lint
         run: make lint
 
-  # TODO: Replace with fetching a released binary once bink publishes GitHub
-  # releases.
-  build-bink:
+  fetch-bink:
```
jlebon: I think if it's just a curl, there's no need to make it a separate job. Let's just make it part of `e2e`. That should save some cycles.

jlebon (approved):

PR #34, opened 2026-05-27
Set the bootc node status

alicefr:
Implementation of milestone 4b, where the daemon sets the bootc node status based on the bootc status --json. The bootc status is executed by the daemon by entering in the pid and mount namespace of process wit pid one with:
```
nsenter --target 1 --mount --pid --setuid 0 --setgid 0 --env -- bootc status --json
```


alicefr:
@jlebon one thing I'm not very satisfied with the current state is that for the e2e test, I'm not really checking which image is booted. The problem is that for knowing exactly what to expect we need to know which version of bink we are using. It isn't deterministic and I would like to avoid to hard code a bink version. Any idea?

cmd/daemon/main.go
```diff
@@ -31,15 +39,37 @@ func main() {
...
 
+	nodeName := os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		setupLog.Error(fmt.Errorf("NODE_NAME not set"), "NODE_NAME environment variable is required")
+		os.Exit(1)
+	}
+
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
 		Scheme: scheme,
+		Cache: cache.Options{
```
jlebon, line 52: I understand why we need this, but it deserves a comment I think.

internal/bootc/executor.go
```diff
@@ -0,0 +1,43 @@
...
+	cmd := exec.CommandContext(ctx,
+		"nsenter",
+		"--target", "1",
+		"--mount",
+		"--pid",
+		"--setuid", "0",
+		"--setgid", "0",
+		"--env",
+		"--",
+		"bootc", "status", "--json",
```
jlebon: Add `--format-version 1`.

internal/bootc/status.go
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+
```
jlebon, line 2: I feel like we should call out a TODO here to move this to a e.g. `bootc-go` package. See also https://github.com/jlebon/bootc-operator/blob/main/docs/PRD.md#relationship-to-flight-control-flightctl.

internal/daemon/reconciler.go
```diff
@@ -47,7 +51,88 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alpha1.BootcNode) error {
+	data, err := r.Executor.Status(ctx)
+	if err != nil {
+		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
+			Type:               bootcv1alpha1.NodeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             bootcv1alpha1.NodeReasonError,
+			Message:            fmt.Sprintf("failed to get bootc status: %v", err),
+			ObservedGeneration: bn.Generation,
```
jlebon, line 76: Hmm, right this makes me realize we're not setting this in the pool controller code. Filed https://github.com/jlebon/bootc-operator/issues/35 for now.

internal/daemon/reconciler_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+	bootcv1alpha1 "github.com/jlebon/bootc-operator/api/v1alpha1"
+	testutil "github.com/jlebon/bootc-operator/test/util"
+)
+
+const (
+	pollInterval = 200 * time.Millisecond
+	pollTimeout  = 10 * time.Second
+
+	testImageRef = "quay.io/example/myos@sha256:aaa111"
```
jlebon: There's a bunch of fake image digests and refs in crd_test.go. We could move those to some place shared.

internal/daemon/reconciler_test.go
```diff
@@ -0,0 +1,241 @@
...
+		g.Expect(got.Status.Conditions).To(ContainElement(And(
+			HaveField("Type", bootcv1alpha1.NodeDegraded),
+			HaveField("Status", metav1.ConditionTrue),
+			HaveField("Reason", bootcv1alpha1.NodeReasonError),
+			HaveField("Message", ContainSubstring("parse")),
+		)))
+	}).Should(Succeed())
+}
+
+func TestReconcileStagedAndRollback(t *testing.T) {
```
jlebon: I would just merge this into `TestReconcilePopulatesStatus`.

internal/daemon/reconciler_test.go
```diff
@@ -0,0 +1,241 @@
...
+	ctx := context.Background()
+
+	fake.set([]byte(bootcStatusBooted), nil)
+
+	// Reconcile a non-existent BootcNode — the reconciler should not
+	// error. We verify indirectly: if the reconciler panicked or
+	// returned a hard error the manager would log it and the test
+	// suite would fail. A brief Consistently check ensures no object
+	// is created.
+	g.Consistently(func(g Gomega) {
```
jlebon: That's a weird test. I would get rid of it I think. It's implicitly tested by e.g. `TestReconcilePopulatesStatus` where there's a lag between when the envtest manager is started and the test is actually started.

internal/daemon/suite_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+const testNodeName = "test-node"
+
+var (
+	testEnv   *envtest.Environment
+	k8sClient client.Client
+	fake      *fakeExecutor
+)
+
+type fakeExecutor struct {
```
jlebon: We'll probably want to move the fake stuff to a separate `_test.go` file eventually.

jlebon:
> @jlebon one thing I'm not very satisfied with the current state is that for the e2e test, I'm not really checking which image is booted. The problem is that for knowing exactly what to expect we need to know which version of bink we are using. It isn't deterministic and I would like to avoid to hard code a bink version. Any idea?

Yeah, agree. Though I think there's an argument here for the bootc-operator driving the `--node-image` instead of using the bink default (even if it'll be the same for now). So then it's controlled here. That would allow us to check the pullspec but not the exact digest. But we could e.g. `podman inspect` before starting the e2e tests and pass it in as an env var.

There's a larger question though around this which is once we actually start testing rollouts, we need to be able to synthesize updates. This is an age-old problem if you look at e.g. rpm-ostree or bootc test suites. One thing we could do for e2e tests is:
- Before starting tests, we pull the node image at the exact digest into the bink registry. We retag it as e.g. `node:latest`.
- When starting nodes, we inject a systemd service that overrides the origin refspec so it actually thinks that it's following `registry.cluster.local:5000/node:latest`. This should probably be exposed as `bink node add --target-imgref` (similar to `bootc install to-filesystem --target-imgref` switch).
- We can assert that in tests etc...
- To test updates, we can create a e.g. `node2:latest` for example and then update the pullspec. Or for tag-following tests, push it to `node:latest`, etc...

Edit: took that to https://github.com/jlebon/bootc-operator/issues/40

internal/bootc/status.go
alicefr: Do you mean an independent bootc.go package right? I mean, not included in the operator code?

jlebon: Right yeah. This would live at e.g. github.com/bootc-dev/bootc-go

test/util/builders.go
```diff
@@ -12,6 +12,20 @@ import (
 	bootcv1alpha1 "github.com/jlebon/bootc-operator/api/v1alpha1"
 )
 
+const (
```
jlebon, line 15: The `testutil` package is more meant to be shared helpers between the envtests and e2e tests. And so it's a bit weird to have test values there. But fine too for now!

jlebon (approved):
Nice, LGTM!

Let's do the test improvements discussed as follow-ups.

jlebon:
Minor follow-ups in https://github.com/jlebon/bootc-operator/pull/46.

Issue #39, opened 2026-05-27
Add example pools


jlebon:
Devs are going to have those locally for testing. We might as well add some in-tree to be nice and to showcase to users.

jlebon:
We can extend those in https://github.com/jlebon/bootc-operator/tree/41701548ae672456d137e9d576dc8d103fba3913/config/samples.

PR #41, opened 2026-05-27
Address low-complexity issues

### LINKED ISSUE #33: Use `--label` to clean up e2e code once it's available
I.e. once https://github.com/alicefr/bink/issues/23 is implemented, use it in the e2e code added in #29.

jlebon:
Closes #33, #36, #38, #39.

### DIFF
--- a/config/samples/bootc_v1alpha1_bootcnodepool.yaml
+++ b/config/samples/bootc_v1alpha1_bootcnodepool.yaml
@@ -1,9 +1,21 @@
+# This sample pool targets all worker nodes. All optional fields are shown
+# commented out with default or example values.
 apiVersion: node.bootc.dev/v1alpha1
 kind: BootcNodePool
 metadata:
-  labels:
-    app.kubernetes.io/name: bootc-operator
-    app.kubernetes.io/managed-by: kustomize
-  name: bootcnodepool-sample
+  name: workers
 spec:
-  # TODO(user): Add fields here
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/worker: ""
+  image:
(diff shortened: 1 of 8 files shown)


config/samples/bootc_v1alpha1_bootcnodepool.yaml
```diff
@@ -1,9 +1,21 @@
...
-    app.kubernetes.io/managed-by: kustomize
-  name: bootcnodepool-sample
+  name: workers
 spec:
-  # TODO(user): Add fields here
+  nodeSelector:
+    matchLabels:
+      node-role.kubernetes.io/worker: ""
+  image:
+    ref: ghcr.io/alicefr/bink/node:latest
```
alicefr, line 12: nit: we still don't support the tag, we should reference the digest

alicefr (approved):

config/samples/bootc_v1alpha1_bootcnodepool.yaml
jlebon: Yeah. Would be nice to have a sample that works here. But the bink/node image has no fixed tag anyway so there's no digest to use. I think I'll just add a comment there for now to say that tags are not supported yet and you'd use a digest.

jlebon:
Merged as part of https://github.com/jlebon/bootc-operator/pull/42. Will address https://github.com/jlebon/bootc-operator/pull/41#discussion_r3316785317 in a follow-up.

config/samples/bootc_v1alpha1_bootcnodepool.yaml
jlebon: Filed https://github.com/jlebon/bootc-operator/issues/45

PR #49, opened 2026-06-03
Prep patches for rollout state machine

jlebon:
Move reboot slot annotations to BootcNode, split rollout code into its own file, and add reboot slot tracking with candidate selection. No transitions yet, just classifying and selecting.

See individual commits for more details.


internal/controller/rollout.go
```diff
@@ -5,33 +5,179 @@ package controller
...
+// rolloutState holds the classified BootcNodes for a single reconcile
+// pass.
+type rolloutState struct {
+	// nodes are sorted into these buckets
+	idle         []*bootcv1alpha1.BootcNode
+	staging      []*bootcv1alpha1.BootcNode
+	staged       []*bootcv1alpha1.BootcNode
+	rebooting    []*bootcv1alpha1.BootcNode
+	degraded     []*bootcv1alpha1.BootcNode
+	unclassified int // nodes we failed to classify
```
alicefr: why not a list in this case as well?

internal/controller/rollout.go
```diff
@@ -5,33 +5,179 @@ package controller
...
+		"idle", len(rs.idle),
+		"staging", len(rs.staging),
+		"staged", len(rs.staged),
+		"rebooting", len(rs.rebooting),
+		"degraded", len(rs.degraded),
+		"unclassified", rs.unclassified,
+		"occupiedSlots", rs.occupiedSlots,
+		"maxUnavailable", maxUnavail,
+		"availableSlots", avail,
+		"candidates", nodeNames(candidates),
```
alicefr, line 67: we could also defined them as additional printable columns. In this way, they are displayed by `kubectl get` . This can be done in a follow up PR

internal/controller/rollout.go
```diff
@@ -5,33 +5,179 @@ package controller
...
+// pool's rollout spec. Defaults to 1 when unset. A value of 0 is allowed and
+// means no reboot slots are available (effectively paused). Returns an
+// invalidSpecError if the value is malformed.
+func resolveMaxUnavailable(pool *bootcv1alpha1.BootcNodePool, nodeCount int) (int, error) {
+	if pool.Spec.Rollout != nil && pool.Spec.Rollout.Paused {
+		return 0, nil
+	}
+	if pool.Spec.Rollout == nil || pool.Spec.Rollout.MaxUnavailable == nil {
+		return 1, nil
+	}
```
alicefr, line 118: nit: switch with the previous if so you can avoid to recheck  `pool.Spec.Rollout != nil`

internal/controller/rollout.go
```diff
@@ -5,33 +5,179 @@ package controller
...
+
+	if len(result) == 0 {
+		return nil
+	}
+	return result
+}
+
+// hasAnnotation returns whether the object has the given annotation key.
+func hasAnnotation(bn *bootcv1alpha1.BootcNode, key string) bool {
```
alicefr: nit: You can use the function `HasAnnotation` from the package `k8s.io/apimachinery/pkg/apis/meta/v1`

alicefr (approved):

internal/controller/rollout.go
jlebon: Because we don't actually need the objects themselves for anything other than the total count. But... we should probably log those anyway, so switched to that!

internal/controller/rollout.go
jlebon: https://github.com/jlebon/bootc-operator/issues/52

internal/controller/rollout.go
jlebon: Not quite equivalent. If `MaxUnavailable` is undefined but `Paused` is true, we should still return 0.

jlebon:
Updated for comments! They're minor changes so will merge based on approval.

PR #50, opened 2026-06-03
Add state machine for image switch (milestone 4c)

alicefr:
Implements the daemon-side state machine that detects image mismatches between spec.desiredImage and the booted image,      
  stages updates via bootc switch in a background goroutine, and triggers reboot when desiredImageState == Booted.            
                                                                                                                            
  - Add Switch(ctx, image, apply) to the Executor interface                                                                   
  - Rewrite the reconciler with async staging, cancel-on-spec-change, and single-patch-per-reconcile                        
  - Background goroutine signals completion via source.Channel, letting the reconcile loop handle all status updates through  
  one path                                                                                                                    
  - Apply operations (reboot) are never cancelled — the daemon restarts after reboot and reconciles fresh

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,6 +68,7 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 		BINK_LOCAL_REGISTRY_NODE_IMAGE=$(BINK_LOCAL_REGISTRY_NODE_IMAGE) \
 		ARTIFACTS=$(ARTIFACTS) \
 		BINK_NODE_IMAGE_DIGEST=$$(skopeo inspect --tls-verify=false --format '{{.Digest}}' docker://localhost:5000/node:latest) \
+		BINK_NODE_IMAGE_UPDATE_DIGEST=$$(skopeo inspect --tls-verify=false docker://localhost:5000/node:update | jq -r '.Digest') \
 		go test -timeout 10m -count=1 $(if $(V),-v) $(if $(RUN),-run $(RUN)) .
 
 ##@ Build
--- a/config/daemon/daemon.yaml
+++ b/config/daemon/daemon.yaml
@@ -38,7 +38,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 512Mi
           requests:
             cpu: 10m
             memory: 64Mi
--- a/internal/bootc/executor.go
+++ b/internal/bootc/executor.go
@@ -6,6 +6,9 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"strings"
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 // Executor abstracts the execution of bootc commands on the host.
@@ -13,6 +16,8 @@ import (
 // PID namespaces. Tests can provide a fake implementation.
 type Executor interface {
 	Status(ctx context.Context) ([]byte, error)
+	Stage(ctx context.Context, image string) error
+	Reboot(ctx context.Context) error
 }
 
 // HostExecutor runs bootc commands on the host via nsenter.
@@ -23,21 +28,46 @@ func NewHostExecutor() *HostExecutor {
 	return &HostExecutor{}
 }
 
-func (e *HostExecutor) Status(ctx context.Context) ([]byte, error) {
-	cmd := exec.CommandContext(ctx,
-		"nsenter",
+func (e *HostExecutor) nsenterCmd(ctx context.Context, args ...string) *exec.Cmd {
+	base := []string{
 		"--target", "1",
-		"--mount",
-		"--pid",
-		"--setuid", "0",
-		"--setgid", "0",
-		"--env",
-		"--",
-		"bootc", "status", "--json", "--format-version", "1",
-	)
+		"--mount", "--pid",
+		"--setuid", "0", "--setgid", "0",
+		"--env", "--",
+	}
+	return exec.CommandContext(ctx, "nsenter", append(base, args...)...)
+}
+
+func (e *HostExecutor) Status(ctx context.Context) ([]byte, error) {
+	cmd := e.nsenterCmd(ctx, "bootc", "status", "--json", "--format-version", "1")
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("running bootc status: %w", err)
 	}
 	return out, nil
+}
+
+func (e *HostExecutor) Stage(ctx context.Context, image string) error {
+	log := logf.FromContext(ctx)
+
+	// TODO: use --download-only once available (https://github.com/bootc-dev/bootc/issues/2137)
+	cmd := e.nsenterCmd(ctx, "bootc", "switch", image)
+	log.Info("Executing", "cmd", strings.Join(cmd.Args, " "))
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("running bootc switch: %s: %w", out, err)
+	}
+	return nil
+}
+
+func (e *HostExecutor) Reboot(ctx context.Context) error {
+	log := logf.FromContext(ctx)
+
+	cmd := e.nsenterCmd(ctx, "systemctl", "reboot")
+	log.Info("Executing", "cmd", strings.Join(cmd.Args, " "))
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("running systemctl reboot: %s: %w", out, err)
+	}
+	return nil
 }
--- a/internal/daemon/fake_test.go
+++ b/internal/daemon/fake_test.go
@@ -4,32 +4,129 @@ package daemon
 
 import (
 	"context"
+	"encoding/json"
+	"strings"
 	"sync"
+
+	testutil "github.com/jlebon/bootc-operator/test/util"
+
+	"github.com/jlebon/bootc-operator/internal/bootc"
 )
 
 type fakeExecutor struct {
-	mu   sync.Mutex
-	data []byte
-	err  error
+	mu        sync.Mutex
+	status    bootc.Status
+	statusErr error
+
+	stageErr  error
+	stageImg  string
+	stageHook func()
+
+	rebootCalled bool
+	rebootErr    error
 }
 
 func (f *fakeExecutor) Status(_ context.Context) ([]byte, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	return f.data, f.err
+	if f.statusErr != nil {
+		return nil, f.statusErr
+	}
+	data, err := json.Marshal(f.status)
+	if err != nil {
+		return nil, err
+	}
+	return data, nil
 }
 
-func (f *fakeExecutor) Stage(_ context.Context, _ string) error {
+func (f *fakeExecutor) Stage(_ context.Context, image string) error {
+	f.mu.Lock()
+	f.stageImg = image
+	f.mu.Unlock()
+
+	if f.stageHook != nil {
+		f.stageHook()
+	}
+	if f.stageErr != nil {
+		return f.stageErr
+	}
+
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	_, digest, _ := strings.Cut(image, "@")
+	f.status.Status.Staged = newBootEntry(image, digest)
 	return nil
 }
 
 func (f *fakeExecutor) Reboot(_ context.Context) error {
-	return nil
-}
-
-func (f *fakeExecutor) set(data []byte, err error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	f.data = data
-	f.err = err
+	f.rebootCalled = true
+	return f.rebootErr
+}
+
+func (f *fakeExecutor) setStatusErr(err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.statusErr = err
+}
+
+func (f *fakeExecutor) setStageErr(err error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.stageErr = err
+}
+
+func (f *fakeExecutor) setStageHook(hook func()) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.stageHook = hook
+}
+
+func (f *fakeExecutor) getStageImg() string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.stageImg
+}
+
+func (f *fakeExecutor) getRebootCalled() bool {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return f.rebootCalled
+}
+
+func (f *fakeExecutor) reset() {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	f.status = bootc.Status{}
+	f.statusErr = nil
+	f.stageErr = nil
+	f.stageImg = ""
+	f.stageHook = nil
+	f.rebootCalled = false
+	f.rebootErr = nil
+}
+
+func newBootEntry(image, digest string) *bootc.BootEntry {
+	return &bootc.BootEntry{
+		Image: &bootc.ImageStatus{
+			Image:        bootc.ImageReference{Image: image, Transport: "registry"},
+			ImageDigest:  digest,
+			Architecture: "amd64",
+		},
+	}
+}
+
+func newBootcStatus(bootedDigest string) bootc.Status {
+	return bootc.Status{
+		APIVersion: "org.containers.bootc/v1alpha1",
+		Kind:       "BootcHost",
+		Spec: bootc.StatusSpec{
+			Image:     &bootc.ImageReference{Image: testutil.ImageTaggedRef, Transport: "registry"},
+			BootOrder: "default",
+		},
+		Status: bootc.StatusBody{
+			Booted: newBootEntry(testutil.ImageTaggedRef, bootedDigest),
+		},
+	}
 }
--- a/internal/daemon/fake_test.go
+++ b/internal/daemon/fake_test.go
@@ -19,6 +19,14 @@ func (f *fakeExecutor) Status(_ context.Context) ([]byte, error) {
 	return f.data, f.err
 }
 
+func (f *fakeExecutor) Stage(_ context.Context, _ string) error {
+	return nil
+}
+
+func (f *fakeExecutor) Reboot(_ context.Context) error {
+	return nil
+}
+
 func (f *fakeExecutor) set(data []byte, err error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
--- a/internal/daemon/fake_test.go
+++ b/internal/daemon/fake_test.go
@@ -22,8 +22,7 @@ type fakeExecutor struct {
 	stageImg  string
 	stageHook func()
 
-	rebootCalled bool
-	rebootErr    error
+	rebooted bool
 }
 
 func (f *fakeExecutor) Status(_ context.Context) ([]byte, error) {
@@ -61,8 +60,8 @@ func (f *fakeExecutor) Stage(_ context.Context, image string) error {
 func (f *fakeExecutor) Reboot(_ context.Context) error {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	f.rebootCalled = true
-	return f.rebootErr
+	f.rebooted = true
+	return nil
 }
 
 func (f *fakeExecutor) setStatusErr(err error) {
@@ -89,10 +88,10 @@ func (f *fakeExecutor) getStageImg() string {
 	return f.stageImg
 }
 
-func (f *fakeExecutor) getRebootCalled() bool {
+func (f *fakeExecutor) getRebooted() bool {
 	f.mu.Lock()
 	defer f.mu.Unlock()
-	return f.rebootCalled
+	return f.rebooted
 }
 
 func (f *fakeExecutor) reset() {
@@ -103,8 +102,7 @@ func (f *fakeExecutor) reset() {
 	f.stageErr = nil
 	f.stageImg = ""
 	f.stageHook = nil
-	f.rebootCalled = false
-	f.rebootErr = nil
+	f.rebooted = false
 }
 
 func newBootEntry(image, digest string) *bootc.BootEntry {
--- a/internal/daemon/reconciler.go
+++ b/internal/daemon/reconciler.go
@@ -5,32 +5,68 @@ package daemon
 import (
 	"context"
 	"fmt"
+	"github.com/distribution/reference"
+	"reflect"
+	"sync"
+	"time"
 
+	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	bootcv1alpha1 "github.com/jlebon/bootc-operator/api/v1alpha1"
 	"github.com/jlebon/bootc-operator/internal/bootc"
 )
 
+const (
+	stageBackoffMin = 5 * time.Second
+	// caps exponential backoff at 5m20s
+	stageMaxBackoffExponent = 7
+)
+
+// stageOp tracks the state of an in-flight bootc stage operation.
+type stageOp struct {
+	mu sync.Mutex
+	// runMu serializes run() calls so that at most one bootc switch
+	// process is active at a time, even if the previous one is still
+	// exiting after context cancellation.
+	runMu   sync.Mutex
+	image   string
+	cancel  context.CancelFunc
+	err     error
+	retries int
+}
+
 // BootcNodeReconciler reconciles the BootcNode for the node this daemon
-// runs on. It reads bootc status from the host and writes it into the
-// BootcNode's status subresource.
+// runs on. It reads bootc status from the host, detects image mismatches,
+// and drives updates via bootc stage.
 type BootcNodeReconciler struct {
 	client.Client
 	Scheme   *runtime.Scheme
 	NodeName string
 	Executor bootc.Executor
+
+	inflight  stageOp
+	stageDone chan event.GenericEvent
+	// rebootIssued tracks whether a reboot has been issued so classifyAction
+	// can distinguish the Staged→Rebooting.
+	rebootIssued bool
 }
 
 func (r *BootcNodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	r.stageDone = make(chan event.GenericEvent, 1)
+
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&bootcv1alpha1.BootcNode{}).
+		WatchesRawSource(source.Channel(r.stageDone, &handler.EnqueueRequestForObject{})).
 		Named("bootcnode").
 		Complete(r)
 }
@@ -51,49 +87,8 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("fetching BootcNode: %w", err)
 	}
 
-	patch := client.MergeFrom(bn.DeepCopy())
-
-	if err := r.populateStatus(ctx, &bn); err != nil {
-		log.Error(err, "Failed to populate bootc status")
-	}
-
-	if err := r.Status().Patch(ctx, &bn, patch); err != nil {
-		return ctrl.Result{}, fmt.Errorf("patching BootcNode status: %w", err)
-	}
-
-	log.Info("Patched BootcNode status from bootc")
-	return ctrl.Result{}, nil
-}
-
-func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alpha1.BootcNode) error {
-	data, err := r.Executor.Status(ctx)
-	if err != nil {
-		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
-			Type:               bootcv1alpha1.NodeDegraded,
-			Status:             metav1.ConditionTrue,
-			Reason:             bootcv1alpha1.NodeReasonError,
-			Message:            fmt.Sprintf("failed to get bootc status: %v", err),
-			ObservedGeneration: bn.Generation,
-		})
-		return fmt.Errorf("getting bootc status: %w", err)
-	}
-
-	status, err := bootc.ParseStatus(data)
-	if err != nil {
-		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
-			Type:               bootcv1alpha1.NodeDegraded,
-			Status:             metav1.ConditionTrue,
-			Reason:             bootcv1alpha1.NodeReasonError,
-			Message:            fmt.Sprintf("failed to parse bootc status: %v", err),
-			ObservedGeneration: bn.Generation,
-		})
-		return fmt.Errorf("parsing bootc status: %w", err)
-	}
-
+	orig := bn.DeepCopy()
 	bn.Status.ObservedGeneration = bn.Generation
-	bn.Status.Booted = convertBootEntry(status.Status.Booted)
-	bn.Status.Staged = convertBootEntry(status.Status.Staged)
-	bn.Status.Rollback = convertBootEntry(status.Status.Rollback)
 
 	apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
 		Type:               bootcv1alpha1.NodeIdle,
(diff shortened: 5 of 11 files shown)


alicefr:
@jlebon I skipped the e2e test since we still need to define how to test the upgrade. The state machine is only checked by the env tests for now

internal/daemon/fake_test.go
```diff
@@ -8,9 +8,13 @@ import (
...
-	mu   sync.Mutex
-	data []byte
-	err  error
+	mu          sync.Mutex
+	data        []byte
+	err         error
+	switchErr   error
+	switchImg   string
+	switchApply bool
+	switchHook  func()
```
jlebon: This is fine for now, though it would be much cleaner if we had something a little higher-fidelity. Basically we can have a `status Status` member and `Status()` serializes that back, and e.g. `Switch()` mutates `Status`, etc.

internal/daemon/reconciler.go
```diff
@@ -51,66 +69,239 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("fetching BootcNode: %w", err)
 	}
 
+	specChanged := bn.Generation > bn.Status.ObservedGeneration
 	patch := client.MergeFrom(bn.DeepCopy())
+	statusOrig := bn.Status.DeepCopy()
```
jlebon: Hmm, we've already done a deep copy of the whole `BootcNode` for the patch above. Since we don't/shouldn't refetch the `BootcNode`, we know the spec will be unchanged, so we can instead move the deep copy to a separate variable and then compare against that to know if to patch?

internal/daemon/reconciler.go
```diff
@@ -51,66 +69,239 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 			Message:            fmt.Sprintf("failed to get bootc status: %v", err),
 			ObservedGeneration: bn.Generation,
 		})
-		return fmt.Errorf("getting bootc status: %w", err)
+		return fmt.Errorf("populating bootc fields: %w", err)
 	}
 
-	status, err := bootc.ParseStatus(data)
-	if err != nil {
+	if bn.Status.Booted == nil {
```
jlebon, line 155: Shouldn't we mark as Degraded here?

internal/daemon/reconciler.go
```diff
@@ -5,32 +5,50 @@ package daemon
...
+// and drives updates via bootc switch.
 type BootcNodeReconciler struct {
 	client.Client
 	Scheme   *runtime.Scheme
 	NodeName string
 	Executor bootc.Executor
+
+	mu           sync.Mutex
+	switchImage  string
+	switchApply  bool
```
jlebon: I _think_ we discussed this at some point. My inclination is to not try to optimize this. I.e. let's not support an `--apply` special-case at all. We just do the `bootc switch` and then use the `GenericEvent` to immediately do another reconcile if `desiredImageState` is `booted`.

internal/daemon/reconciler.go
```diff
@@ -51,66 +69,239 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
 			Type:               bootcv1alpha1.NodeDegraded,
 			Status:             metav1.ConditionTrue,
 			Reason:             bootcv1alpha1.NodeReasonError,
-			Message:            fmt.Sprintf("failed to parse bootc status: %v", err),
+			Message:            fmt.Sprintf("bootc switch failed: %v", switchErr),
 			ObservedGeneration: bn.Generation,
 		})
-		return fmt.Errorf("parsing bootc status: %w", err)
+		return nil
```
jlebon: Hmm even if we encounter a switch error, we should keep retrying the operation, in K8s fashion.

internal/daemon/reconciler.go
```diff
@@ -51,66 +69,239 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			Status:             metav1.ConditionFalse,
+			Reason:             bootcv1alpha1.NodeReasonStaged,
+			ObservedGeneration: bn.Generation,
+		})
+		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
+			Type:               bootcv1alpha1.NodeDegraded,
+			Status:             metav1.ConditionFalse,
+			Reason:             bootcv1alpha1.NodeReasonHealthy,
+			ObservedGeneration: bn.Generation,
+		})
```
jlebon: Why do we restore healthy here but not in the other cases?

jlebon: To dedupe a bit, we could just set a `reason` variable in the switch cases and then only call `SetStatusCondition()` once before returning.

internal/bootc/executor.go
```diff
@@ -23,21 +24,36 @@ func NewHostExecutor() *HostExecutor {
...
+func (e *HostExecutor) Status(ctx context.Context) ([]byte, error) {
+	cmd := e.nsenterCmd(ctx, "bootc", "status", "--json", "--format-version", "1")
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("running bootc status: %w", err)
 	}
 	return out, nil
 }
+
+func (e *HostExecutor) Switch(ctx context.Context, image string, apply bool) error {
```
jlebon: Let's perhaps add a docstring here to use `--download-only` once it's available and link to https://github.com/bootc-dev/bootc/issues/2137.

internal/daemon/reconciler.go
```diff
@@ -5,32 +5,50 @@ package daemon
...
 	NodeName string
 	Executor bootc.Executor
+
+	mu           sync.Mutex
+	switchImage  string
+	switchApply  bool
+	switchCancel context.CancelFunc
+	switchErr    error
+
+	switchDone chan event.GenericEvent
```
jlebon: I think one thing that would make the code cleaner is to group this into an e.g. `BootcStagingOp` struct and then we can have convenience receiver methods to manipulate the group.

internal/daemon/reconciler.go
```diff
@@ -51,66 +69,239 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		log.Info("Cancelling in-flight switch", "old", r.switchImage, "new", desiredImage)
+		r.switchCancel()
+		r.switchImage = ""
+		r.switchApply = false
+		r.switchCancel = nil
+	}
+	return false
+}
+
+func (r *BootcNodeReconciler) runSwitch(ctx context.Context, image string, apply bool) {
```
jlebon: I think a docstring here that mentions this runs in a separate goroutine would be helpful.

internal/daemon/reconciler_test.go
```diff
@@ -20,7 +21,38 @@ const (
...
 
-	testImageRef = testutil.ImageDigestRefA
+	switchErrMsg      = "switch failed: pull error"
+	bootcStatusErrMsg = "bootc status failed"
+
+	testImageRef      = testutil.ImageDigestRefA
+	testOtherImageRef = testutil.ImageDigestRefB
+	testThirdImageRef = testutil.ImageDigestRefC
+
+	bootcStatusBooted = `{
```
jlebon: Yeah, these status literals are... not great. Let's try to do the fake executor enhancement I mentioned earlier.

internal/daemon/reconciler.go
alicefr: I wasn't sure, I thought the same but then decide to just retrigger the reconciliation loop. But probably it is a good idea to mark that the node is in a weird state

internal/daemon/reconciler.go
alicefr: To clarify, ` --apply` is not used during staging, we always stage with plain bootc switch <image>. The `--apply` is only used in the reboot step: when the image is already staged and desiredImageState is Booted, we run `bootc switch --apply` to boot into it.

Are you suggesting we should decouple this further and trigger the reboot through a different mechanism (e.g. `systemctl reboot`) instead of `bootc switch --apply`?

internal/daemon/reconciler.go
alicefr: I added some delay if there are any transitioning error

internal/daemon/reconciler.go
alicefr: Sorry, this is wrong bootc switch --apply doesn't reboot, but it is instead `bootc upgrade --apply`, the right command

alicefr:
There are still some issue with this PR... investigating.

alicefr:
I think we should prioritize https://github.com/jlebon/bootc-operator/pull/57 and implement at least the happy path test for the reboot

internal/bootc/executor.go
```diff
@@ -23,21 +28,45 @@ func NewHostExecutor() *HostExecutor {
...
 	if err != nil {
 		return nil, fmt.Errorf("running bootc status: %w", err)
 	}
 	return out, nil
 }
+
+func (e *HostExecutor) Switch(ctx context.Context, image string) error {
+	log := logf.FromContext(ctx)
+
+	cmd := e.nsenterCmd(ctx, "bootc", "switch", image)
```
jlebon, line 54: TODO link to https://github.com/bootc-dev/bootc/issues/2137 would still be good here.

config/daemon/daemon.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         resources:
           limits:
             cpu: 500m
-            memory: 128Mi
+            memory: 512Mi
```
jlebon, line 41: The commit message needs a rationale.

internal/daemon/fake_test.go
```diff
@@ -19,6 +19,14 @@ func (f *fakeExecutor) Status(_ context.Context) ([]byte, error) {
 	return f.data, f.err
 }
 
+func (f *fakeExecutor) Switch(_ context.Context, _ string) error {
+	return nil
+}
+
+func (f *fakeExecutor) Upgrade(_ context.Context) error {
```
jlebon: Shouldn't this be Reboot instead?

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("fetching BootcNode: %w", err)
 	}
 
-	patch := client.MergeFrom(bn.DeepCopy())
+	specChanged := bn.Generation > bn.Status.ObservedGeneration
+	orig := bn.DeepCopy()
+	patch := client.MergeFrom(orig)
```
jlebon: Do we even need a `patch` variable? I would just inline the `client.MergeFrom(orig)` in the `Patch()` call.

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+func (r *BootcNodeReconciler) reconcileBootcNode(ctx context.Context, bn *bootcv1alpha1.BootcNode, specChanged bool) (ctrl.Result, error) {
+	log := logf.FromContext(ctx).WithValues("node", r.NodeName)
+
+	if err := r.populateBootcFields(ctx, bn); err != nil {
+		log.Error(err, "Failed to populate bootc status")
 		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
 			Type:               bootcv1alpha1.NodeDegraded,
 			Status:             metav1.ConditionTrue,
 			Reason:             bootcv1alpha1.NodeReasonError,
 			Message:            fmt.Sprintf("failed to get bootc status: %v", err),
```
jlebon: OK as is, but I wonder if it would be cleaner to instead define a new error type, e.g. `degradedError` which gets "caught" in the top-level `Reconcile()` loop and sets `NodeReasonError` and the message to from the degradedError object. See also `invalidSpecError()` in the controller code.

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 
 	apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
 		Type:               bootcv1alpha1.NodeIdle,
-		Status:             metav1.ConditionTrue,
-		Reason:             bootcv1alpha1.NodeReasonIdle,
+		Status:             metav1.ConditionFalse,
+		Reason:             reason,
 		ObservedGeneration: bn.Generation,
 	})
 	apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
```
jlebon: One pattern I found useful in the controller code is to upfront set the initial condition to "no errors"; e.g. here in the Reconcile code after we deepcopy, we set degraded to false. And then code in the whole reconciliation loop only needs to care about setting degraded true as it comes upon error conditions (which with the previous comment would be shortened to just returning a `degradedError`).

Actually, this probably generalizes to NodeIdle too. E.g. set NodeIdle to true upfront and then only code that needs to set it to false needs to write it out.

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
+			Type:               bootcv1alpha1.NodeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             bootcv1alpha1.NodeReasonError,
+			Message:            fmt.Sprintf("bootc switch failed: %v", switchErr),
+			ObservedGeneration: bn.Generation,
+		})
+		// Requeue with a delay to retry transient failures (e.g. network
+		// blips, registry timeouts) without hammering the registry.
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
jlebon: Wonder if exponential backoff is worth it here, up to some cap probably. E.g. we don't want to back off to multiple hours. But I could imagine then we could start from e.g. 5s and increase until... 5 mins perhaps?

internal/daemon/reconciler.go
```diff
@@ -108,9 +204,99 @@ func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alp
...
+	bn.Status.Rollback = convertBootEntry(status.Status.Rollback)
+
 	return nil
 }
 
+// imageNeedsUpdate compares only the digest portion of desiredImage against
+// bootedDigest. It assumes upgrades always come from the same image repository.
+// TODO: also compare the image repository to detect cross-image switches.
+func imageNeedsUpdate(desiredImage, bootedDigest string) bool {
+	_, digest, ok := strings.Cut(desiredImage, "@")
```
jlebon: First, why do we even need to cut here? `bootc status --json` already has a field with just the digest.

Second, if we really wanted to work from a digested pullspec, we should parse it as a proper OCI ref here. See e.g. `parseImageRef()` in the controller.

But I think the proper fix is the first one.

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+
+	reboot := bn.Spec.DesiredImageState == bootcv1alpha1.DesiredImageStateBooted
+	desiredImage := bn.Spec.DesiredImage
+
+	if skip := r.inflight.acquire(log, desiredImage); skip {
+		return ctrl.Result{}, nil
+	}
+
+	_, desiredDigest, _ := strings.Cut(desiredImage, "@")
```
jlebon: Yeah, should parse here.

internal/daemon/reconciler.go
```diff
@@ -108,9 +204,99 @@ func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alp
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	err := s.err
+	s.err = nil
+	return err
+}
+
+// acquire checks whether a switch operation is already in flight.
+// It returns true if the reconciler should skip (operation in progress),
+// or false after cancelling any stale in-flight switch.
```
jlebon: Minor: I would expect the opposite semantics here: `true` means we successfully acquired the switchOp, and `false` means we didn't.

internal/daemon/reconciler.go
```diff
@@ -51,54 +73,128 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	var reason string
+	switch {
+	case !alreadyStaged:
+		reason = bootcv1alpha1.NodeReasonStaging
+		switchCtx, cancel := context.WithCancel(context.Background())
+		r.inflight.start(desiredImage, cancel)
+		log.Info("Starting staging", "image", desiredImage)
+		go r.inflight.run(switchCtx, r.NodeName, desiredImage, r.Executor, r.switchDone)
+
+	case reboot && idleCond != nil && (idleCond.Reason == bootcv1alpha1.NodeReasonStaged || idleCond.Reason == bootcv1alpha1.NodeReasonRebooting):
```
jlebon: Hmm, I don't fully understand this.

I'd like to try to structure this switch in a way that's easier to read and tell that it's right.

I wonder if something similar to the node state classification we do in the controller would help here. E.g. have an "enum" of the current state of the node purely based on its `bootc status`, and then classify upfront what state we're in, and then the final switch here which switches on the state.

internal/daemon/reconciler.go
```diff
@@ -108,9 +204,99 @@ func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alp
...
+	if s.cancel != nil {
+		log.Info("Cancelling in-flight switch", "old", s.image, "new", image)
+		s.cancel()
+		s.image = ""
+		s.cancel = nil
+	}
+	return false
+}
+
+func (s *switchOp) start(image string, cancel context.CancelFunc) {
```
jlebon: `start` is quite similar to `run`. Perhaps `initialize` or `prepare`?

internal/daemon/reconciler.go
```diff
@@ -108,9 +204,99 @@ func (r *BootcNodeReconciler) populateStatus(ctx context.Context, bn *bootcv1alp
...
+	s.err = nil
+}
+
+func (s *switchOp) run(ctx context.Context, nodeName, image string, executor bootc.Executor, done chan<- event.GenericEvent) {
+	log := logf.FromContext(context.Background()).WithValues("node", nodeName, "image", image)
+
+	err := executor.Switch(ctx, image)
+
+	s.mu.Lock()
+	if ctx.Err() != nil {
```
jlebon, line 287: I guess one TODO here is to switch to exec'ing the `bootc switch` async, and then a `select` here where we monitor the cancel channel so we can actually send a SIGINT to bootc.

jlebon:
I haven't re-reviewed past the 2nd commit yet (tests).

internal/daemon/reconciler.go
jlebon: > The --apply is only used in the reboot step: when the image is already staged and desiredImageState is Booted, we run bootc switch --apply to boot into it.

OK, that makes sense and I guess it does match closer to what we'll actually do once we have `--from-downloaded`.

I do like the executor verb being called `Reboot()` though as you have now to make this clearer. So maybe then we just change it to take an image arg so it runs `bootc switch --apply`.

But yeah, I think the bug you're hitting is https://github.com/bootc-dev/bootc/pull/2249.

internal/bootc/executor.go
alicefr: Should I rename the function `Stage` then? It doesn't follow bootc API but it makes the intention more explicit like the reboot

jlebon: WFM!

internal/daemon/reconciler.go
alicefr: Actually, we can treat all the error from the reconcileBootcNode as Degraded errors and we don't need an extra type

jlebon: Hmm, maybe. I think it's helpful to distinguish between "OS errors", and any other error (e.g. failed to get or patch some k8s object). I think only the former should cause a `NodeDegraded` condition.

alicefr: The `reconcileBootcNode` does already return an error only if there are an `OS error`. Except if we we consider the `bn.Status.Booted == nil ` not as an OS error. The other cases is when a bootc operation fails.

internal/daemon/reconciler.go
alicefr: I'm not sure I understand the comment. The desired image is from the spec not the status. Hence, it might not be present in the status if not once staged

internal/daemon/reconciler.go
alicefr: Do you mean to use `parseImageRef` ?

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, fmt.Errorf("fetching BootcNode: %w", err)
 	}
 
-	patch := client.MergeFrom(bn.DeepCopy())
+	specChanged := bn.Generation > bn.Status.ObservedGeneration
```
jlebon: This seems unused.

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	orig := bn.DeepCopy()
+	bn.Status.ObservedGeneration = bn.Generation
+
+	apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
```
jlebon, line 93: Perhaps a comment here similar to https://github.com/bootc-dev/bootc-operator/blob/3a27f6e9056a8a7b0c150d9bd4c0dca853ebc80e/internal/controller/bootcnodepool_controller.go#L228?

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
+			Type:               bootcv1alpha1.NodeDegraded,
+			Status:             metav1.ConditionTrue,
+			Reason:             bootcv1alpha1.NodeReasonError,
+			Message:            reconcileErr.Error(),
+			ObservedGeneration: bn.Generation,
+		})
+		if result.RequeueAfter > 0 {
+			reconcileErr = nil
+		}
```
jlebon: Interesting. This is worth a comment perhaps. Returning an error also causes a requeue AFAIK.

So what happens if both a RequeueAfter is set _and_ we have a non-nil error? I assume based on the fact you had to add this code that it doesn't do what we want. But I'm curious why exactly.

internal/daemon/reconciler.go
jlebon: Got it, makes sense. Hmm, debating whether the BootcNode spec should implicitly enforce digests by having separate `DesiredImage` and `DesiredImageDigest` fields... Anyway, no need to block on that for now.

WDYT about upfront in `reconcileBootcNode()` we parse `desiredImage` using `parseImageRef()` and then we pass the `reference.Named` object around as needed? We could also then error out early if somehow it's not a digested pullspec. And it should also make it easier to resolve the TODO in the docstring here (pass the bootedImage as well and parse it too and then you can compare its `ref.Name()` to our desired `ref.Name()`).

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if err != nil {
+		s.retries++
+	}
+	return err
+}
+
+func (s *stageOp) backoff() time.Duration {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	d := stageBackoffMin << (s.retries - 1)
```
jlebon: `s.retries` is not bounded AFAICT, so this will keep shifting until we overflow and then d will be back at 0, no?

Simplest is to also cap `retries` in the if branch below.

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	if !r.inflight.acquire(log, desiredImage) {
+		return ctrl.Result{}, nil
 	}
 
-	log.Info("Patched BootcNode status from bootc")
+	action := classifyAction(bn)
+
+	var reason string
+	switch action {
```
jlebon, line 199: Nice, love how much easier it is to read now!

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		reason = bootcv1alpha1.NodeReasonStaging
+		switchCtx, cancel := context.WithCancel(context.Background())
+		r.inflight.initialize(desiredImage, cancel)
+		log.Info("Starting staging", "image", desiredImage)
+		go r.inflight.run(switchCtx, r.NodeName, desiredImage, r.Executor, r.stageDone)
+
+	case actionReboot:
+		reason = bootcv1alpha1.NodeReasonRebooting
+		log.Info("Starting reboot", "image", desiredImage)
+		if err := r.Executor.Reboot(ctx); err != nil {
```
jlebon: There's a chance here that we get killed before we even reach the `Patch()` to update our conditions. Perhaps instead we should return another bool from this function to say when a reboot is necessary, and then `Reconcile()` itself can trigger it at the very end?

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	return digest != bootedDigest
+}
+
+// updateAction represents the next step the daemon should take for an
+// in-progress update. Classified once, then used to drive the stage.
+type updateAction int
+
+const (
+	actionStage       updateAction = iota // desired image not yet staged
+	actionAwaitStaged                     // staged, waiting for reboot approval
```
jlebon: Perhaps `actionAwaitBooted`?

internal/daemon/reconciler.go
```diff
@@ -51,66 +79,253 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+
+	switch idleCond.Reason {
+	case bootcv1alpha1.NodeReasonStaged:
+		return actionReboot
+	case bootcv1alpha1.NodeReasonRebooting:
+		return actionAwaitReboot
+	default:
+		return actionAwaitStaged
+	}
```
jlebon: Hmm, I don't understand why we need to inspect our own conditions here instead of deriving from truth.

How about we extend e.g. `BootcNodeReconciler` with a `rebootIssued bool`, and then here we can just check that to know if to return `actionAwaitReboot` or `actionReboot`?

internal/daemon/reconciler.go
alicefr: if you set the error, the re-queue delay is ignored and it uses the default one. It is there to implement our back-off when bootc stage fails. But I agree it isn't that great.

alicefr: One option is that we separate the degraded message from the error. When we want to exponentially back-off we don't return the error and we set the delay for the requeue. I was merging the errors, but it is probably better to separate the message from when we want to error during the reconcile

internal/daemon/reconciler.go
alicefr: It was difficult to derive and distinguish if we are in staged or in rebooting without checking the condition. If the desiredImageState is boot. In this way, we first patch the condition with staged, this re trigger a reconciliation loop and then we set the rebooting

internal/daemon/reconciler.go
alicefr: In the end, I decided to move the degraded condition in the if else statement where it handled after the reconcile evaluation. Mainly, for 2 reason. 

First, the same condition is handled in the same statement. 

Secondly, it avoids the flipping of the `LastTransitionTimestamp` set it by SetStatusCondition. If we set first always the healthy path and then it is degraded, we flip back the condition on multiple reconciliation loops. Then, the DeepEqual function detects that the status has changed and always patch the conditions because the `LastTransitionTimestamp` has changed.

internal/daemon/reconciler.go
alicefr: Set the max number of retries

internal/daemon/reconciler.go
jlebon: > Secondly, it avoids the flipping of the `LastTransitionTimestamp` set it by SetStatusCondition.

Hmm, yeah that's a good point. I guess that's a bug in the bootcnodepool controller as well we should fix. Tracking as https://github.com/bootc-dev/bootc-operator/issues/66 for now.

internal/daemon/reconciler.go
```diff
@@ -5,32 +5,65 @@ package daemon
...
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	bootcv1alpha1 "github.com/jlebon/bootc-operator/api/v1alpha1"
 	"github.com/jlebon/bootc-operator/internal/bootc"
 )
 
+const (
+	stageBackoffMin = 5 * time.Second
+	// caps exponential backoff at 5m20s
+	stageBackoffMaxRetries = 7
```
jlebon: Minor/optional: hmm, naming is misleading here. Perhaps `stageMaxBackoffExponent`?

internal/daemon/reconciler.go
```diff
@@ -5,32 +5,65 @@ package daemon
...
 	client.Client
 	Scheme   *runtime.Scheme
 	NodeName string
 	Executor bootc.Executor
+
+	inflight  stageOp
+	stageDone chan event.GenericEvent
+	// rebootIssued tracks whether a reboot has been issued so classifyAction
+	// can distinguish the Staged→Rebooting transition without reading back
+	// its own status conditions.
```
jlebon: Minor/optional: I would drop the "without reading..."

This is a classic AI thing where during a refactor it adds comments very specific to how the old approach worked as a rationale even though it's not really useful for new readers.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+}
+
+// reconcileBootcNode defines the result of the reconcile of the bootc nodes. It return the results for the reconcile,
+// the degrade message and eventual errors. We distinguish the degraded message from a reconcile error since we want to
+// implement an exponential back-off if the staging failed.
+func (r *BootcNodeReconciler) reconcileBootcNode(ctx context.Context, bn *bootcv1alpha1.BootcNode) (reconcileResult, error) {
+	log := logf.FromContext(ctx).WithValues("node", r.NodeName)
+
+	if err := r.populateBootcFields(ctx, bn); err != nil {
+		degradedErr := fmt.Errorf("failed to get bootc status: %w", err)
```
jlebon: Minor/optional: this duplicates the error context from the callee. Perhaps "populating bootc fields: %w".

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	if bn.Status.Booted == nil {
+		degradedErr := fmt.Errorf("bootc status has no booted entry")
+		return reconcileResult{degradedMsg: degradedErr.Error()}, degradedErr
+	}
+
+	desiredRef, err := reference.ParseNamed(bn.Spec.DesiredImage)
 	if err != nil {
+		degradedErr := fmt.Errorf("invalid image ref %q: %w", bn.Spec.DesiredImage, err)
+		return reconcileResult{degradedMsg: degradedErr.Error()}, degradedErr
```
jlebon: Should return nil here for second element. DesiredImage is simply invalid so there's no point retrying. We'll get triggered anyway when it changes.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	}
+	// The controller always resolves tags to digests at the pool level.
+	digested, ok := desiredRef.(reference.Digested)
+	if !ok {
+		degradedErr := fmt.Errorf("image ref %q has no digest", bn.Spec.DesiredImage)
+		return reconcileResult{degradedMsg: degradedErr.Error()}, degradedErr
```
jlebon: Here as well.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		r.reset()
 		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
-			Type:               bootcv1alpha1.NodeDegraded,
+			Type:               bootcv1alpha1.NodeIdle,
 			Status:             metav1.ConditionTrue,
-			Reason:             bootcv1alpha1.NodeReasonError,
-			Message:            fmt.Sprintf("failed to get bootc status: %v", err),
+			Reason:             bootcv1alpha1.NodeReasonIdle,
 			ObservedGeneration: bn.Generation,
 		})
```
jlebon: Is this necessary? We already default to NodeIdle true in the caller.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+	if stageErr != nil {
 		apimeta.SetStatusCondition(&bn.Status.Conditions, metav1.Condition{
-			Type:               bootcv1alpha1.NodeDegraded,
+			Type:               bootcv1alpha1.NodeIdle,
 			Status:             metav1.ConditionTrue,
-			Reason:             bootcv1alpha1.NodeReasonError,
-			Message:            fmt.Sprintf("failed to parse bootc status: %v", err),
+			Reason:             bootcv1alpha1.NodeReasonIdle,
 			ObservedGeneration: bn.Generation,
 		})
```
jlebon: Here too.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		s.cancel()
+		s.retries = 0
+	}
+	s.image = image
+	s.cancel = cancel
+	s.err = nil
+}
+
+func (r *BootcNodeReconciler) reset() {
+	r.rebootIssued = false
```
jlebon, line 270: Hmm, I don't think we need this? `rebootIssued` should be a one-direction change.

So then this method can be directly on the `stageOp` receiver.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+
+	// TODO: exec bootc switch async and select on the cancel channel to send SIGINT for graceful shutdown.
+	err := executor.Stage(ctx, image)
+
+	s.mu.Lock()
+	if ctx.Err() != nil {
+		log.Info("Stage cancelled")
+	} else {
+		if err != nil {
+			log.Error(err, "Stage failed")
```
jlebon: Might as well include the image in the log line here.

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+			s.err = err
+		}
+		s.image = ""
+		s.cancel = nil
+	}
+	s.mu.Unlock()
+
+	if ctx.Err() == nil {
+		done <- event.GenericEvent{
```
jlebon: Why not do this in the else block above?

internal/daemon/fake_test.go
```diff
@@ -4,32 +4,131 @@ package daemon
...
+	testutil "github.com/jlebon/bootc-operator/test/util"
+
+	"github.com/jlebon/bootc-operator/internal/bootc"
 )
 
 type fakeExecutor struct {
-	mu   sync.Mutex
-	data []byte
-	err  error
+	mu        sync.Mutex
```
jlebon, line 17: Hmm interesting. Why do we need a Mutex here?

internal/daemon/fake_test.go
```diff
@@ -4,32 +4,131 @@ package daemon
...
+	}
+	return data, nil
 }
 
-func (f *fakeExecutor) Stage(_ context.Context, _ string) error {
+func (f *fakeExecutor) Stage(_ context.Context, image string) error {
+	f.mu.Lock()
+	f.stageImg = image
+	hook := f.stageHook
+	err := f.stageErr
```
jlebon: These feel not worth aliasing.

internal/daemon/reconciler_test.go
jlebon: Should use v1 here, and similarly below for v2 and v3.

internal/daemon/reconciler_test.go
```diff
@@ -22,9 +22,12 @@ const (
 	pollInterval = 200 * time.Millisecond
 	pollTimeout  = 10 * time.Second
 
+	stageErrMsg       = "stage failed: pull error"
 	bootcStatusErrMsg = "bootc status failed"
 
-	testImageRef = testutil.ImageDigestRefA
+	testImageRef      = testutil.ImageDigestRefA
+	testOtherImageRef = testutil.ImageDigestRefB
+	testThirdImageRef = testutil.ImageDigestRefC
```
jlebon: Is this worth aliasing? It's a bit longer but not by much. At least one test below is using a mixture of the alias and the originals, which makes it hard to follow. Not aliasing guarantees we're consistent. :)

internal/daemon/reconciler_test.go
```diff
@@ -120,3 +123,201 @@ func TestReconcileBootcStatusError(t *testing.T) {
...
+	t.Cleanup(func() {
+		_ = k8sClient.Delete(ctx, bn)
+	})
+
+	g.Eventually(func() string {
+		return fake.getStageImg()
+	}).Should(Equal(testOtherImageRef))
+
+	fake.setStageHook(nil)
+	close(firstBlock)
```
jlebon, line 307: This mock test made me realize something; we're not actually waiting for an in-flight `switch` operation to finish before issuing a new one, right? We cancel the context, but don't actually wait for the previous goroutine to end.

That's fine I guess, but ideally we should have at most only one `bootc switch` active at a time. In practice, they're effectively serialized because bootc will take the sysroot lock early on and the following one will block, but I also want to proactively guard against a buggy situation where `bootc switch` hangs and we're somehow piling on goroutines all spamming more `bootc switch` calls.

I guess a simple way to handle this is to just add another mutex and hold it for the entire body of `stageOp.run()`?

internal/daemon/reconciler.go
```diff
@@ -51,66 +84,280 @@ func (r *BootcNodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+}
+
+func (r *BootcNodeReconciler) reset() {
+	r.rebootIssued = false
+	r.inflight.mu.Lock()
+	defer r.inflight.mu.Unlock()
+	r.inflight.retries = 0
+}
+
+func (s *stageOp) run(ctx context.Context, nodeName, image string, executor bootc.Executor, done chan<- event.GenericEvent) {
```
jlebon, line 277: Still would be nice to have a comment here that mentions this runs in a goroutine. :)

Makefile
```diff
@@ -68,6 +68,7 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 		BINK_LOCAL_REGISTRY_NODE_IMAGE=$(BINK_LOCAL_REGISTRY_NODE_IMAGE) \
 		ARTIFACTS=$(ARTIFACTS) \
 		BINK_NODE_IMAGE_DIGEST=$$(skopeo inspect --tls-verify=false --format '{{.Digest}}' docker://localhost:5000/node:latest) \
+		UPDATE_IMAGE_DIGEST=$$(skopeo inspect --tls-verify=false docker://localhost:5000/node:update | jq -r '.Digest') \
```
jlebon: `BINK_NODE_IMAGE_UPDATE_DIGEST`?

test/e2e/e2eutil/env.go
```diff
@@ -60,6 +60,10 @@ type Env struct {
 	// nodeImageRegistry is the in-cluster registry path for the seeded node image
 	// (e.g. "registry.cluster.local:5000/node"). Empty when not seeded.
 	nodeImageRegistry string
+
+	// updateImageDigest is the manifest digest of the update image
+	// (e.g. "sha256:def456..."). Empty when not built.
+	updateImageDigest string
```
jlebon: And here and in the new fields and functions too e.g. `nodeImageUpdateDigest`/`NodeImageUpdateDigest()`.

config/daemon/daemon.yaml
jlebon, line 41: OK wow, yeah. The rationale for this is a great reason why we should switch to `systemd-run` to invoke the switch. This is something I wanted to do already, but yeah, it's now further motivated.

Don't need to switch it here, but let's definitely file an issue for it so we remember that follow up.

test/e2e/e2eutil/env.go
```diff
@@ -80,6 +84,7 @@ func New(t *testing.T) *Env {
 
 	nodeImageDigest := os.Getenv("BINK_NODE_IMAGE_DIGEST")
 	nodeImageRegistry := os.Getenv("BINK_LOCAL_REGISTRY_NODE_IMAGE")
+	updateImageDigest := os.Getenv("UPDATE_IMAGE_DIGEST")
```
jlebon: I feel like we should just `t.Fatal` here if any of those are unset. That way, individual tests don't also need to check this.

test/e2e/bootcnode_test.go
```diff
@@ -87,3 +90,114 @@ func TestControllerMembership(t *testing.T) {
...
+		)))
+	}).WithTimeout(3 * time.Minute).Should(Succeed())
+
+	t.Logf("Node %q is Idle with original image", nodeName)
+
+	// Phase 2: Patch pool to update image.
+	updateRef := env.UpdateImageDigestedPullSpec()
+	if updateRef == "" {
+		t.Fatal("UPDATE_IMAGE_DIGEST must be set")
+	}
```
jlebon: E.g. we could drop this then.

test/e2e/bootcnode_test.go
```diff
@@ -87,3 +90,114 @@ func TestControllerMembership(t *testing.T) {
...
+
+	modified := pool.DeepCopy()
+	modified.Spec.Image.Ref = updateRef
+	g.Expect(env.Client.Patch(ctx, modified, client.MergeFrom(pool))).To(Succeed())
+	*pool = *modified
+
+	t.Logf("Patched pool to update image %s", updateRef)
+
+	// Phase 3: Wait for Rebooting — proves image was staged and reboot started.
+	// The Staged phase is tested separately with rollout paused.
```
jlebon: Is this talking about a future test that's not part of this PR?

jlebon (commented):
Full review this time.

This is getting quite close! Very cool to see it all tested in the e2e test.

config/daemon/daemon.yaml
alicefr: Do we want to? I mean at least we're accounting the memory consumption in what it is actually consuming the memory.

I do recall that we want to switch to `systemd-run` to not kill the bootc switch operation eventually on an upgrade of the daemon. But I also like the idea that we could get the OOM kill in kubernetes while if this is happening at system level we won't know unless we are monitoring it.

internal/daemon/reconciler.go
alicefr: Yeah, sorry I missed this

internal/daemon/reconciler.go
alicefr: I think we need it to reset the reboot because if we change the desired image. It seems that the image as already reboot it. We need to clean the state for the next staging flow.

internal/daemon/reconciler.go
alicefr: It is already included in the log message from the [line](https://github.com/alicefr/bootc-operator/blob/milestone-4c/internal/daemon/reconciler.go#L281) above 
```golang
log := logf.FromContext(context.Background()).WithValues("node", nodeName, "image", image)
```

internal/daemon/fake_test.go
alicefr: Because we run the fake Stage in a goroutine, so it access the `fakeExecutor` fields concurrently with the test. For example the test [TestCancelInflightStage](https://github.com/alicefr/bootc-operator/blob/milestone-4c/internal/daemon/reconciler_test.go#L286C6-L286C29) calls Stage while the test goroutine polls `fake.getStageImg()`. Both touches the `stageImg` concurrently

test/e2e/bootcnode_test.go
alicefr: yes, I can remove it, if this is misleading

internal/daemon/reconciler_test.go
alicefr: do you mean another switch operation for the same image?

test/e2e/bootcnode_test.go
alicefr: I added as a TODO

internal/daemon/reconciler_test.go
alicefr: I guess you meant 2 separate images before the previous one was cancelled, we could have 2 switching operations running in parallel. Better to be on the safe side.

alicefr:
There has been some flaky test runs. I have removed the check for the rebooting reason because apparently the bootcnode is sometimes set to degraded if during reboot the bootc status doesn't report correctly the status. It was hard to detect it because the daemon restarts and we loose the logs while the controller was reporting only degraded without the reason and the message. 

Although, the code is correct, even if the bootcnode is set to degraded for some time, it then recovers since the test failure was reporting healthy condition. Example of a run can be found [here](https://github.com/bootc-dev/bootc-operator/actions/runs/27823822405/job/82346496821).

The e2e test verifies now that the end state is the desired one and we successfully managed to upgrade without checking intermediate conditions.

internal/daemon/reconciler.go
jlebon: If the desired image changed, but we already issued a reboot for a previous image, it's too late to change course. The node will go down at any moment.

alicefr: But what if the reboot continue to fail and we want to change the image.

jlebon: If the machine is failing to reboot, then we got bigger problems. :)

I think once we initiate a reboot, the node should be considered in an undefined state, and trying to make changes (if we're somehow still alive) is going to have unreliable results.

In fact, we should match what the MCO does and just very early on in the reconcile check if a reboot has been issued and `return nil`: https://github.com/openshift/machine-config-operator/blob/ae127c3c06eddd12a7f801a34f84109317ab0cbc/pkg/daemon/daemon.go#L745-L750

config/daemon/daemon.yaml
jlebon: > But I also like the idea that we could get the OOM kill in kubernetes while if this is happening at system level we won't know unless we are monitoring it.

If bootc switch gets oomkilled, then the switch operation would fail, and that would cause the BootcNode to degrade, no? It wouldn't be as obvious that it's an oomkill, but at least we have visibility that something is wrong.

But more conceptually, I think of it as two separate layers: the K8s layer and the OS layer. We run in the Kubernetes layer and yes, drive OS layer bits, but the OS layer needs to be well-configured and stand on its own. It shouldn't be in scope for us to have to determine what a reasonable amount of memory a bootc switch should take. That feels like an abstraction violation of sorts.

But anyway, to be clear, don't need to change anything here. I'll file an issue for it and we can discuss more there!

internal/daemon/reconciler.go
jlebon: This is not a blocker/can be a follow-up.

jlebon:
> There has been some flaky test runs. I have removed the check for the rebooting reason because apparently the bootcnode is sometimes set to degraded if during reboot the bootc status doesn't report correctly the status.

Nice. Yeah this would be fixed by https://github.com/bootc-dev/bootc-operator/pull/50#discussion_r3443296157.

internal/daemon/reconciler_test.go
```diff
@@ -114,7 +115,205 @@ func TestReconcileBootcStatusError(t *testing.T) {
...
+			HaveField("Type", bootcv1alpha1.NodeIdle),
+			HaveField("Status", metav1.ConditionFalse),
+			HaveField("Reason", bootcv1alpha1.NodeReasonRebooting),
+		)))
+	}).Should(Succeed())
+
+	g.Expect(fake.getRebooted()).To(BeTrue())
+}
+
+func TestRollback(t *testing.T) {
```
jlebon, line 251: Another test similar to this one would be a e.g. `TestUnstaged`: booted is A, staged is B, and then we create a BN with desired=A.

And hmm, right. What _ideally_ happens is that we unstage B. But what will actually happen is that we "switch to A" and create another staged copy of A. bootc will happily do this it seems enough it's a no-op. But we should still ideally handle this nicely.

Anyway, obviously non-blocking.

jlebon (approved):
LGTM, nice work! 🎉

internal/daemon/reconciler_test.go
jlebon: This is actually slightly more serious than I thought. What will actually happen is that we no-op very early because booted == desired, and so we never even get to the "switch to A" bit and having bootc create a copy of A. That means that until we have [finalization locking](https://github.com/bootc-dev/bootc/issues/2137), any reboot in that state will reboot the node into B.

jlebon: Anyway, leaning towards leaving this as is and not handling this for now. Once we do have locking this mostly becomes cosmetic, and it doesn't seem worth trying to fix those kinds of issues just yet.

config/daemon/daemon.yaml
alicefr: >If bootc switch gets oomkilled, then the switch operation would fail, and that would cause the BootcNode to degrade, no? It wouldn't be as obvious that it's an oomkill, but at least we have visibility that something is wrong.

If this isn't run by the daemon then we need to monitor the systemd unit right? I don't think bootc status reports that a staging operation failed

> But more conceptually, I think of it as two separate layers: the K8s layer and the OS layer. We run in the Kubernetes layer and yes, drive OS layer bits, but the OS layer needs to be well-configured and stand on its own. It shouldn't be in scope for us to have to determine what a reasonable amount of memory a bootc switch should take. That feels like an abstraction violation of sorts.

IMO, it belongs to our layer. For example, if we are downloading the image, the node might get busy. If you don't report the memory that the pod could consume, it is harder for the k8s scheduler know how to schedule pods. 
There are also different class [quality of service](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/) in k8s, I think we should set our to [Guaranteed](https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#guaranteed)

PR #57, opened 2026-06-09
Set the target image reference for the bink node

alicefr:
Overwrite the node image reported by `bootc status` to the image pushed in the local registry. Also, it infers the digest of the node image from the label stored on the node disk image.

Additionally, this PR prepares a second image for the testing of the update. It generates a second image starting from the node image and push it to the local registry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,12 @@ build-daemon: ## Build daemon binary.
 buildimg: ## Build container image.
 	$(CONTAINER_TOOL) build -t $(IMG) .
 
+.PHONY: build-update-image
+build-update-image: ## Build a derived node image for update testing and push to bink registry.
+	@printf 'FROM localhost:5000/node:latest\nRUN touch /usr/share/update-marker\n' | \
+		podman build -t localhost:5000/node:update -f - .
+	podman push --tls-verify=false localhost:5000/node:update
+
 ##@ Deployment
 
 ifndef ignore-not-found
@@ -135,7 +141,7 @@ start-bink: seed-node-image ## Start a bink cluster (idempotent).
 	kubectl --kubeconfig $(KUBECONFIG_BINK) wait --for=condition=Ready node/controller --timeout=5m
 
 .PHONY: deploy-bink
-deploy-bink: start-bink kustomize ## Deploy to a bink cluster (requires: buildimg).
+deploy-bink: start-bink build-update-image kustomize ## Deploy to a bink cluster (requires: buildimg).
 	podman push --tls-verify=false $(IMG) localhost:5000/bootc-operator-e2e:latest
 	# On re-deploy, restart the rollout to force a re-pull of the :latest tag.
 	# On fresh deploy, skip the restart -- the pod is already pulling the correct image.
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,8 @@ CONTAINER_TOOL ?= podman
 BINK_CLUSTER_NAME ?= e2e
 KUBECONFIG_BINK ?= ./kubeconfig-$(BINK_CLUSTER_NAME)
 ARTIFACTS ?= $(abspath _output/logs)
+BINK_NODE_DISK_IMAGE ?= ghcr.io/alicefr/bink/node:v1.35-fedora-44-disk
+BINK_LOCAL_REGISTRY_NODE_IMAGE ?= registry.cluster.local:5000/node
 # YEAR defines the year value used for substituting the YEAR placeholder in the boilerplate header.
 YEAR ?= $(shell date +%Y)
 
@@ -62,7 +64,10 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 	rm -rf $(ARTIFACTS)
 	cd test/e2e && KUBECONFIG=$(abspath $(KUBECONFIG_BINK)) BINK_CLUSTER_NAME=$(BINK_CLUSTER_NAME) \
 		$(if $(BINK_NODE_IMAGE),BINK_NODE_IMAGE=$(BINK_NODE_IMAGE)) \
+		BINK_NODE_DISK_IMAGE=$(BINK_NODE_DISK_IMAGE) \
+		BINK_LOCAL_REGISTRY_NODE_IMAGE=$(BINK_LOCAL_REGISTRY_NODE_IMAGE) \
 		ARTIFACTS=$(ARTIFACTS) \
+		BINK_NODE_IMAGE_DIGEST=$$(skopeo inspect --tls-verify=false --format '{{.Digest}}' docker://localhost:5000/node:latest) \
 		go test -timeout 10m -count=1 $(if $(V),-v) $(if $(RUN),-run $(RUN)) .
 
 ##@ Build
@@ -111,12 +116,22 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
 # Note the :latest tag here: this makes the pull policy be Always.
 IMG_BINK ?= registry.cluster.local:5000/bootc-operator-e2e:latest
 
-.PHONY: start-bink
-start-bink: ## Start a bink cluster (idempotent).
+.PHONY: seed-node-image
+seed-node-image: ## Pull the bootc node image by digest and push to the bink registry.
 	bink registry start
-	bink cluster list 2>&1 | grep -qw $(BINK_CLUSTER_NAME) || \
+	podman pull $(BINK_NODE_DISK_IMAGE)
+	bootc_img=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		bootc_digest=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		podman pull "$$bootc_img@$$bootc_digest" && \
+		podman tag "$$bootc_img@$$bootc_digest" localhost:5000/node:latest
+	podman push --tls-verify=false localhost:5000/node:latest
+
+.PHONY: start-bink
+start-bink: seed-node-image ## Start a bink cluster (idempotent).
+	bink cluster list 2>&1 | grep -qw $(BINK_CLUSTER_NAME) || { \
+		node_digest=$$(skopeo inspect --tls-verify=false --format '{{.Digest}}' docker://localhost:5000/node:latest) && \
 		bink cluster start --cluster-name $(BINK_CLUSTER_NAME) --node-name controller --api-port 0 --expose $(KUBECONFIG_BINK) \
-		$(if $(BINK_NODE_IMAGE),--node-image $(BINK_NODE_IMAGE))
+		--node-image $(BINK_NODE_DISK_IMAGE) --target-imgref $(BINK_LOCAL_REGISTRY_NODE_IMAGE)@$$node_digest; }
 	kubectl --kubeconfig $(KUBECONFIG_BINK) wait --for=condition=Ready node/controller --timeout=5m
 
 .PHONY: deploy-bink
--- a/test/e2e/bootcnode_test.go
+++ b/test/e2e/bootcnode_test.go
@@ -35,8 +35,12 @@ func TestControllerMembership(t *testing.T) {
 
 	ctx := context.Background()
 
-	// Create a pool selecting this test's nodes.
+	// Use real registry image when seeded; fall back to fake ref for
+	// local dev runs without deploy-bink.
 	imageRef := testutil.ImageDigestRefA
+	if ref := env.NodeImageDigestedPullSpec(); ref != "" {
+		imageRef = ref
+	}
 	pool := env.NewPool("workers", imageRef)
 	g.Expect(env.Client.Create(ctx, pool)).To(Succeed())
 
--- a/test/e2e/bootcnode_test.go
+++ b/test/e2e/bootcnode_test.go
@@ -14,7 +14,6 @@ import (
 
 	bootcv1alpha1 "github.com/jlebon/bootc-operator/api/v1alpha1"
 	"github.com/jlebon/bootc-operator/test/e2e/e2eutil"
-	testutil "github.com/jlebon/bootc-operator/test/util"
 )
 
 const (
@@ -35,13 +34,7 @@ func TestControllerMembership(t *testing.T) {
 
 	ctx := context.Background()
 
-	// Use real registry image when seeded; fall back to fake ref for
-	// local dev runs without deploy-bink.
-	imageRef := testutil.ImageDigestRefA
-	if ref := env.NodeImageDigestedPullSpec(); ref != "" {
-		imageRef = ref
-	}
-	pool := env.NewPool("workers", imageRef)
+	pool := env.NewPool("workers", env.NodeImageDigestedPullSpec())
 	g.Expect(env.Client.Create(ctx, pool)).To(Succeed())
 
 	// Wait for BootcNode to appear for the worker.
@@ -56,7 +49,7 @@ func TestControllerMembership(t *testing.T) {
 	g.Expect(owner.Name).To(Equal(pool.Name))
 
 	// Verify desiredImage.
-	g.Expect(bn.Spec.DesiredImage).To(Equal(imageRef))
+	g.Expect(bn.Spec.DesiredImage).To(Equal(env.NodeImageDigestedPullSpec()))
 
 	// Verify the worker has the managed label.
 	var node corev1.Node
@@ -83,8 +76,10 @@ func TestControllerMembership(t *testing.T) {
 	g.Eventually(func(g Gomega) {
 		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
 		g.Expect(bn.Status.Booted).NotTo(BeNil(), "expected booted status to be populated")
-		g.Expect(bn.Status.Booted.Image).NotTo(BeEmpty(), "expected booted image to be non-empty")
-		g.Expect(bn.Status.Booted.ImageDigest).NotTo(BeEmpty(), "expected booted imageDigest to be non-empty")
+		g.Expect(bn.Status.Booted.Image).To(Equal(env.NodeImageDigestedPullSpec()),
+			"booted image should match seeded registry image")
+		g.Expect(bn.Status.Booted.ImageDigest).To(Equal(env.NodeImageDigest()),
+			"booted image digest should match seeded registry image")
 		g.Expect(bn.Status.Conditions).To(ContainElement(And(
 			HaveField("Type", bootcv1alpha1.NodeIdle),
 			HaveField("Status", metav1.ConditionTrue),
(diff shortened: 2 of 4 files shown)


alicefr:
We still have some mismatch with the digest. So, it is expected that the test failed, but the PR is ready for review. The problem is on bink

test/e2e/e2eutil/env.go
```diff
@@ -51,6 +51,14 @@ type Env struct {
...
 	// nodes tracks node names added via AddNode for cleanup.
 	nodes []string
+
+	// nodeImageDigest is the manifest digest of the bootc image seeded
+	// into the bink registry (e.g. "sha256:abc123..."). Empty when not seeded.
+	nodeImageDigest string
+
+	// nodeImgRegistry is the in-cluster registry path for the seeded
+	// node image (e.g. "registry.cluster.local:5000/node"). Empty when not seeded.
+	nodeImgRegistry string
```
jlebon: Hmm, `s/Img/Image/` for consistency.

Makefile
```diff
@@ -110,10 +114,17 @@ IMG_BINK ?= registry.cluster.local:5000/bootc-operator:e2e
 .PHONY: deploy-bink
 deploy-bink: kustomize ## Deploy to a bink cluster (idempotent).
 	bink registry start
+	podman pull $(BINK_NODE_DISK_IMAGE)
+	bootc_img=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		podman pull "$$bootc_img" && \
+		podman tag "$$bootc_img" localhost:5000/node:latest
+	podman push --tls-verify=false localhost:5000/node:latest
```
jlebon: How about something like `skopeo copy --preserve-digests "$$bootc_img" localhost:5000/node:latest`. `--preserve-digests` ensures that the digests remain the same (or fail).

Another huge benefit would be that we don't actually have to pay the cost of the node image in the local containers-storage. But I guess we do need it for now to be able to build the derived node image. Will see if we can avoid that in a follow-up…

Makefile
```diff
@@ -110,10 +114,17 @@ IMG_BINK ?= registry.cluster.local:5000/bootc-operator:e2e
...
+	podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE) > .node-digest
 	podman push --tls-verify=false $(IMG) localhost:5000/bootc-operator:e2e
 	bink cluster list 2>&1 | grep -qw $(BINK_CLUSTER_NAME) || \
 		bink cluster start --cluster-name $(BINK_CLUSTER_NAME) --node-name controller --api-port 0 --expose $(KUBECONFIG_BINK) \
-		$(if $(BINK_NODE_IMAGE),--node-image $(BINK_NODE_IMAGE))
+		$(if $(BINK_NODE_DISK_IMAGE),--node-image $(BINK_NODE_DISK_IMAGE)) \
+		--target-imgref $(NODE_IMG_REGISTRY):latest
```
jlebon: Feels weird to have an env var for this but not for the localhost:5000 version. Either we don't have for either or we have for both.

Makefile
```diff
@@ -59,7 +61,9 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 # each package, even though we just have one here--but I really like streaming
 # output...).
 	cd test/e2e && KUBECONFIG=$(abspath $(KUBECONFIG_BINK)) BINK_CLUSTER_NAME=$(BINK_CLUSTER_NAME) \
-		$(if $(BINK_NODE_IMAGE),BINK_NODE_IMAGE=$(BINK_NODE_IMAGE)) \
+		$(if $(BINK_NODE_DISK_IMAGE),BINK_NODE_DISK_IMAGE=$(BINK_NODE_DISK_IMAGE)) \
+		$(if $(wildcard .node-digest),NODE_IMAGE_DIGEST=$$(cat $(abspath .node-digest))) \
```
jlebon: Instead of the `.node-digest`, couldn't we just query the `BINK_NODE_DISK_IMAGE` for the digest here instead?

Makefile
```diff
@@ -110,10 +114,17 @@ IMG_BINK ?= registry.cluster.local:5000/bootc-operator:e2e
...
+	bootc_img=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		podman pull "$$bootc_img" && \
+		podman tag "$$bootc_img" localhost:5000/node:latest
+	podman push --tls-verify=false localhost:5000/node:latest
+	podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE) > .node-digest
 	podman push --tls-verify=false $(IMG) localhost:5000/bootc-operator:e2e
 	bink cluster list 2>&1 | grep -qw $(BINK_CLUSTER_NAME) || \
 		bink cluster start --cluster-name $(BINK_CLUSTER_NAME) --node-name controller --api-port 0 --expose $(KUBECONFIG_BINK) \
-		$(if $(BINK_NODE_IMAGE),--node-image $(BINK_NODE_IMAGE))
+		$(if $(BINK_NODE_DISK_IMAGE),--node-image $(BINK_NODE_DISK_IMAGE)) \
```
jlebon: This no longer needs to be `if` since now that variable is always defined.

Makefile
```diff
@@ -5,6 +5,8 @@ CONTAINER_TOOL ?= podman
 # To use a separate dev cluster: make deploy-bink BINK_CLUSTER_NAME=dev
 BINK_CLUSTER_NAME ?= e2e
 KUBECONFIG_BINK ?= ./kubeconfig-$(BINK_CLUSTER_NAME)
+BINK_NODE_DISK_IMAGE ?= ghcr.io/alicefr/bink/node:v1.35-fedora-44-disk
+NODE_IMG_REGISTRY ?= registry.cluster.local:5000/node
```
jlebon: Should have a `BINK_` prefix. How about something like `BINK_LOCAL_REGISTRY_NODE_IMAGE`?

Makefile
```diff
@@ -59,7 +61,9 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 # each package, even though we just have one here--but I really like streaming
 # output...).
 	cd test/e2e && KUBECONFIG=$(abspath $(KUBECONFIG_BINK)) BINK_CLUSTER_NAME=$(BINK_CLUSTER_NAME) \
-		$(if $(BINK_NODE_IMAGE),BINK_NODE_IMAGE=$(BINK_NODE_IMAGE)) \
+		$(if $(BINK_NODE_DISK_IMAGE),BINK_NODE_DISK_IMAGE=$(BINK_NODE_DISK_IMAGE)) \
```
jlebon: Should be able to drop this `if` too now that it's always set.

Makefile
```diff
@@ -59,7 +61,9 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
...
+		$(if $(wildcard .node-digest),NODE_IMAGE_DIGEST=$$(cat $(abspath .node-digest))) \
+		$(if $(NODE_IMG_REGISTRY),NODE_IMG_REGISTRY=$(NODE_IMG_REGISTRY)) \
```
jlebon: Here too, we can drop the `if` I think.

Makefile
jlebon: I guess we might as well just `export` these variables so then there's no need to even explicitly have to pass them through to the e2e harness.

test/e2e/e2eutil/env.go
```diff
@@ -121,6 +144,10 @@ func (e *Env) AddNode(t *testing.T, opts ...NodeOption) string {
 		o(cfg)
 	}
 
+	if cfg.targetImgRef == "" && e.nodeImgRegistry != "" {
```
jlebon: Let's just hard require this. We control everything here so let's keep the code simple.

(This is related to the "BYO node image" dev story, and I do want to provide a good UX there, but we can work through those details separately.)

Makefile
jlebon: And let's prefix the env vars even in the e2e process with `BINK_`.

test/e2e/e2eutil/env.go
```diff
@@ -168,6 +198,15 @@ func (e *Env) TestLabels() map[string]string {
 	return map[string]string{LabelE2ETest: e.testID}
 }
 
+// NodeImageDigestRef returns the digest-qualified reference for the
+// seeded node image (e.g. "registry.cluster.local:5000/node@sha256:abc123").
+func (e *Env) NodeImageDigestRef() string {
```
jlebon: I tend to call this the "digested pull spec". Perhaps `NodeImageDigestedPullSpec`?

test/e2e/bootcnode_test.go
```diff
@@ -35,8 +35,12 @@ func TestControllerMembership(t *testing.T) {
 
 	ctx := context.Background()
 
-	// Create a pool selecting this test's nodes.
+	// Use real registry image when seeded; fall back to fake ref for
+	// local dev runs without deploy-bink.
 	imageRef := testutil.ImageDigestRefA
+	if ref := env.NodeImageDigestRef(); ref != "" {
```
jlebon: Yeah this is a good example, I want our tests to not have these kinds of if conditions that weaken what they can test.

Makefile
```diff
@@ -83,6 +83,12 @@ build-daemon: ## Build daemon binary.
 buildimg: ## Build container image.
 	$(CONTAINER_TOOL) build -t $(IMG) .
 
+.PHONY: build-update-image
+build-update-image: ## Build a derived node image for update testing and push to bink registry.
+	@printf 'FROM localhost:5000/node:latest\nRUN touch /update-marker\n' | \
```
jlebon: Let's put the marker in e.g. `/usr/share/`?

test/e2e/bootcnode_test.go
```diff
@@ -85,6 +85,17 @@ func TestControllerMembership(t *testing.T) {
 		g.Expect(bn.Status.Booted).NotTo(BeNil(), "expected booted status to be populated")
 		g.Expect(bn.Status.Booted.Image).NotTo(BeEmpty(), "expected booted image to be non-empty")
 		g.Expect(bn.Status.Booted.ImageDigest).NotTo(BeEmpty(), "expected booted imageDigest to be non-empty")
+	}).WithTimeout(3 * time.Minute).Should(Succeed())
+
+	if env.NodeImageDigest() != "" {
+		g.Expect(bn.Status.Booted.Image).To(Equal(env.NodeImageTaggedRef()),
```
jlebon: This is fine for now, but note that how this will actually work is that bootc only ever sees/receives digested pull specs, so there's never going to be a tag.

The `--target-imgref` we pass should always be a digested pull spec I think.

Makefile
alicefr: yes, right now we need to pull it anyway for the image for the updates

test/e2e/bootcnode_test.go
alicefr: Right now, I'm passing the tag not the digest, but let me verify if I can construct the image based on the label of the node disk

alicefr: it works with the digeset. Using the digest instead of the tag

Makefile
```diff
@@ -111,19 +114,28 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
...
-		$(if $(BINK_NODE_IMAGE),--node-image $(BINK_NODE_IMAGE))
+		--node-image $(BINK_NODE_DISK_IMAGE) \
+		--target-imgref $(BINK_LOCAL_REGISTRY_NODE_IMAGE)@$$bootc_digest; }
 	kubectl --kubeconfig $(KUBECONFIG_BINK) wait --for=condition=Ready node/controller --timeout=5m
 
 .PHONY: deploy-bink
 deploy-bink: start-bink kustomize ## Deploy to a bink cluster (requires: buildimg).
 	podman push --tls-verify=false $(IMG) localhost:5000/bootc-operator-e2e:latest
-	# On re-deploy, restart the rollout to force a re-pull of the :latest tag.
-	# On fresh deploy, skip the restart -- the pod is already pulling the correct image.
```
jlebon, line 126: Is there a reason we're dropping this comment?

Makefile
```diff
@@ -111,19 +114,28 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
...
+	bootc_img=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		bootc_digest=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE)) && \
+		podman pull "$$bootc_img@$$bootc_digest" && \
+		podman tag "$$bootc_img@$$bootc_digest" localhost:5000/node:latest
+	podman push --tls-verify=false localhost:5000/node:latest
+
+.PHONY: start-bink
+start-bink: seed-node-image ## Start a bink cluster (idempotent).
+	bink cluster list 2>&1 | grep -qw $(BINK_CLUSTER_NAME) || { \
+		bootc_digest=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE)) && \
```
jlebon: The bootc digest here should come from a `skopeo inspect` on `$BINK_LOCAL_REGISTRY_NODE_IMAGE` to rule out any chance of recompression.

Or alternatively, use `skopeo copy --preserve-digests` in `seed-node-image` instead of `podman push`.

Makefile
```diff
@@ -111,19 +114,28 @@ undeploy: kustomize ## Undeploy controller from the K8s cluster specified in ~/.
...
 		bink cluster start --cluster-name $(BINK_CLUSTER_NAME) --node-name controller --api-port 0 --expose $(KUBECONFIG_BINK) \
-		$(if $(BINK_NODE_IMAGE),--node-image $(BINK_NODE_IMAGE))
+		--node-image $(BINK_NODE_DISK_IMAGE) \
+		--target-imgref $(BINK_LOCAL_REGISTRY_NODE_IMAGE)@$$bootc_digest; }
```
jlebon: Minor/optional: could merge these two lines.

test/e2e/bootcnode_test.go
```diff
@@ -85,6 +78,15 @@ func TestControllerMembership(t *testing.T) {
...
 		g.Expect(bn.Status.Booted.ImageDigest).NotTo(BeEmpty(), "expected booted imageDigest to be non-empty")
+	}).WithTimeout(3 * time.Minute).Should(Succeed())
+
+	g.Expect(bn.Status.Booted.Image).To(Equal(env.NodeImageDigestedPullSpec()),
+		"booted image should match seeded registry image")
+	g.Expect(bn.Status.Booted.ImageDigest).To(Equal(env.NodeImageDigest()),
+		"booted image digest should match seeded registry image")
+
+	g.Eventually(func(g Gomega) {
+		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
```
jlebon: Minor/optional: do we actually need to re-get here?

Makefile
```diff
@@ -85,6 +85,12 @@ build-daemon: ## Build daemon binary.
 buildimg: ## Build container image.
 	$(CONTAINER_TOOL) build -t $(IMG) .
 
+.PHONY: build-update-image
+build-update-image: ## Build a derived node image for update testing and push to bink registry.
+	@printf 'FROM localhost:5000/node:latest\nRUN touch /usr/share/update-marker\n' | \
+		podman build -t localhost:5000/node:update -f - .
+	podman push --tls-verify=false localhost:5000/node:update
```
jlebon, line 94: Minor/optional: for future tests to be able to use this, we should add another `podman inspect` on this in the e2e target to pass through the digest as a variable. Happy to leave as a follow-up for when tests will need this.

Makefile
```diff
@@ -135,6 +141,7 @@ start-bink: seed-node-image ## Start a bink cluster (idempotent).
 
 .PHONY: deploy-bink
 deploy-bink: start-bink kustomize ## Deploy to a bink cluster (requires: buildimg).
+	$(MAKE) build-update-image
```
jlebon: Minor/optional: Was there a reason why we're calling it this way instead of just having it be a target dependency? We might've discussed this, but I don't remember the details...

Makefile
```diff
@@ -6,6 +6,8 @@ CONTAINER_TOOL ?= podman
 BINK_CLUSTER_NAME ?= e2e
 KUBECONFIG_BINK ?= ./kubeconfig-$(BINK_CLUSTER_NAME)
 ARTIFACTS ?= $(abspath _output/logs)
+export BINK_NODE_DISK_IMAGE ?= ghcr.io/alicefr/bink/node:v1.35-fedora-44-disk
+export BINK_LOCAL_REGISTRY_NODE_IMAGE ?= registry.cluster.local:5000/node
```
jlebon: Sorry for the flip-flop here, I know I suggested the export here but upon seeing the code and thinking more about this, I don't like how it's inconsistent with the other vars here (some of which we manually pass through to commands and some which are Makefile local). It's repetitive, but I guess there's value in passing it only on the commands that actually need it because then it's clear who the users really are.

Makefile
```diff
@@ -63,6 +65,7 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 	cd test/e2e && KUBECONFIG=$(abspath $(KUBECONFIG_BINK)) BINK_CLUSTER_NAME=$(BINK_CLUSTER_NAME) \
 		$(if $(BINK_NODE_IMAGE),BINK_NODE_IMAGE=$(BINK_NODE_IMAGE)) \
 		ARTIFACTS=$(ARTIFACTS) \
+		NODE_IMAGE_DIGEST=$$(podman inspect --format '{{index .Config.Labels "bink.bootc-image-digest"}}' $(BINK_NODE_DISK_IMAGE)) \
```
jlebon: Can we call this `BINK_NODE_IMAGE_DIGEST`?

Also, let's `skopeo inspect` on `BINK_LOCAL_REGISTRY_NODE_IMAGE` instead here as well.

jlebon (approved):
Mostly minor comments; LGTM overall!

I filed #60 for the Justfile migration, but probably also eventually we should move a lot of the bink setup stuff to a dedicated `hack/` script file to make it more manageable.

Feel free to merge after fixing at least the non-optional comments.

test/e2e/bootcnode_test.go
alicefr: yes, actually we can merge it with the previous eventually, thanks!

Makefile
alicefr: let's leave it for the real implementation

alicefr: This is part of #50 and there is a dedicated commit  (https://github.com/jlebon/bootc-operator/pull/50/commits/7543a5c813d495703d5dafd534dbc69113d6a05b)

jlebon (approved):
LGTM!

PR #58, opened 2026-06-10
controller: post-reboot handling and deployedDigest

jlebon:
Add the key missing part in the rollout logic: we now actually free reboot slots once nodes reboot into the desired image and are Ready. This allows the state machine to continue rolling out to other nodes. We also keep the `deployedDigest` and `updateAvailable` fields up to date in the pool status.

With this I think we've more or less reached MVP level for the controller part at least. :tada: Most of the remaining work is more about richer reporting back to the pool status and strengthening error-handling.

### DIFF
--- a/internal/controller/rollout_envtest_test.go
+++ b/internal/controller/rollout_envtest_test.go
@@ -27,10 +27,10 @@ func TestSimpleRollout(t *testing.T) {
 
 	const (
 		poolName = "rollout-3node"
-		// All nodes are booting digest B; pool targets digest A.
-		targetRef   = testImageDigestRefA
-		targetImage = testImageDigestRefA
-		bootedImage = testImageDigestRefB
+		// All nodes are booted on digest A; pool targets digest B.
+		oldImage    = testImageDigestRefA
+		newImage    = testImageDigestRefB
+		newImageRef = testImageDigestRefB
 	)
 
 	// Create 3 worker nodes.
@@ -45,7 +45,7 @@ func TestSimpleRollout(t *testing.T) {
 	}
 
 	// Create pool targeting digest A with maxUnavailable: 1.
-	pool := testutil.NewPool(poolName, targetRef,
+	pool := testutil.NewPool(poolName, newImageRef,
 		testutil.WithWorkerSelector(),
 		testutil.WithMaxUnavailable(intstr.FromInt32(1)),
 	)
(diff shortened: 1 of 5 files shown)


internal/controller/rollout_envtest_test.go
```diff
@@ -27,10 +27,10 @@ func TestSimpleRollout(t *testing.T) {
 
 	const (
 		poolName = "rollout-3node"
-		// All nodes are booting digest B; pool targets digest A.
-		targetRef   = testImageDigestRefA
-		targetImage = testImageDigestRefA
-		bootedImage = testImageDigestRefB
+		// All nodes are booted on digest A; pool targets digest B.
+		oldImage    = testImageDigestRefA
```
alicefr, line 33: +1

internal/controller/rollout_envtest_test.go
```diff
@@ -145,3 +156,22 @@ func simulateDaemonStatus(g Gomega, ctx context.Context, nodeName, bootedDigest,
...
+	g.Expect(k8sClient.Get(ctx, client.ObjectKey{Name: nodeName}, &node)).To(Succeed())
+
+	node.Status.Conditions = []corev1.NodeCondition{
+		{
+			Type:               corev1.NodeReady,
+			Status:             corev1.ConditionTrue,
+			LastHeartbeatTime:  metav1.Now(),
+			LastTransitionTime: metav1.Now(),
+		},
+	}
```
alicefr: Probably not important, but you are overwriting the Conditions if there were some set:
```suggestion 
	node.Status.Conditions = append(node.Status.Conditions, corev1.NodeCondition{
		{
			Type:               corev1.NodeReady,
			Status:             corev1.ConditionTrue,
			LastHeartbeatTime:  metav1.Now(),
			LastTransitionTime: metav1.Now(),
		},
	}
```

alicefr:
One question: deployedDigest is never set if any node is degraded or unclassified, even if the rollout itself completed successfully. Is that what we want? Is it a stuck node avoid the rollout to be completed? I guess this follows the same principle as a stuck node holds its reboot slot

internal/controller/rollout_envtest_test.go
jlebon: Yeah I don't think it matters, but fixed anyway!

jlebon:
> One question: deployedDigest is never set if any node is degraded or unclassified, even if the rollout itself completed successfully. Is that what we want? Is it a stuck node avoid the rollout to be completed? I guess this follows the same principle as a stuck node holds its reboot slot

Yeah, thanks for raising this. deployedDigest in general I'm not quite satisfied with to be frank, and I think we should probably drop it. I filed https://github.com/jlebon/bootc-operator/issues/62 for that and for now dropped that commit from this PR.

alicefr (approved):

PR #59, opened 2026-06-11
 Detect external bootc status changes via fsnotify (milestone 4d)

alicefr:
Bootc changes are detected via fsnotify or via a polling mechanism.

### DIFF
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -14,6 +15,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
 	bootcv1alpha1 "github.com/bootc-dev/bootc-operator/api/v1alpha1"
@@ -32,6 +34,9 @@ func init() {
 }
 
 func main() {
+	var pollInterval time.Duration
+	flag.DurationVar(&pollInterval, "bootc-poll-interval", 5*time.Minute, "Interval for polling bootc status as a fallback to fsnotify")
+
 	opts := zap.Options{
 		Development: true,
 	}
@@ -62,17 +67,33 @@ func main() {
 		os.Exit(1)
 	}
 
+	executor := bootc.NewHostExecutor()
+
+	watcher := &daemon.StatusWatcher{
+		PollInterval: pollInterval,
+		OstreePath:    daemon.DefaultOstreePath,
+		ComposefsPath: daemon.DefaultComposefsPath,
+		Events:       make(chan event.GenericEvent, 1),
+		NodeName:     nodeName,
+		Executor:     executor,
+	}
+	if err := mgr.Add(watcher); err != nil {
+		setupLog.Error(err, "Failed to add status watcher")
+		os.Exit(1)
+	}
+
 	if err := (&daemon.BootcNodeReconciler{
-		Client:   mgr.GetClient(),
-		Scheme:   mgr.GetScheme(),
-		NodeName: nodeName,
-		Executor: bootc.NewHostExecutor(),
+		Client:        mgr.GetClient(),
+		Scheme:        mgr.GetScheme(),
+		NodeName:      nodeName,
+		Executor:      executor,
+		StatusWatcher: watcher,
 	}).SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "Failed to create controller", "controller", "bootcnode")
 		os.Exit(1)
 	}
 
-	setupLog.Info("Starting daemon", "node", nodeName)
+	setupLog.Info("Starting daemon", "node", nodeName, "pollInterval", pollInterval)
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		setupLog.Error(err, "Failed to run daemon")
 		os.Exit(1)
--- a/cmd/daemon/main.go
+++ b/cmd/daemon/main.go
@@ -70,12 +70,12 @@ func main() {
 	executor := bootc.NewHostExecutor()
 
 	watcher := &daemon.StatusWatcher{
-		PollInterval: pollInterval,
+		PollInterval:  pollInterval,
 		OstreePath:    daemon.DefaultOstreePath,
 		ComposefsPath: daemon.DefaultComposefsPath,
-		Events:       make(chan event.GenericEvent, 1),
-		NodeName:     nodeName,
-		Executor:     executor,
+		Events:        make(chan event.GenericEvent, 1),
+		NodeName:      nodeName,
+		Executor:      executor,
 	}
 	if err := mgr.Add(watcher); err != nil {
 		setupLog.Error(err, "Failed to add status watcher")
--- a/internal/daemon/reconciler.go
+++ b/internal/daemon/reconciler.go
@@ -46,13 +46,14 @@ type stageOp struct {
 }
 
 // BootcNodeReconciler reconciles the BootcNode for the node this daemon
-// runs on. It reads bootc status from the host, detects image mismatches,
-// and drives updates via bootc stage.
+// runs on. It reads bootc status from the watcher's cache, detects image
+// mismatches, and drives updates via bootc stage.
 type BootcNodeReconciler struct {
 	client.Client
-	Scheme   *runtime.Scheme
-	NodeName string
-	Executor bootc.Executor
+	Scheme        *runtime.Scheme
+	NodeName      string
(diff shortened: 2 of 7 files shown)


internal/daemon/reconciler.go
```diff
@@ -58,17 +58,23 @@ type BootcNodeReconciler struct {
...
 
-	return ctrl.NewControllerManagedBy(mgr).
+	builder := ctrl.NewControllerManagedBy(mgr).
 		For(&bootcv1alpha1.BootcNode{}).
 		WatchesRawSource(source.Channel(r.stageDone, &handler.EnqueueRequestForObject{})).
-		Named("bootcnode").
-		Complete(r)
+		Named("bootcnode")
+
+	if r.StatusChanged != nil {
```
jlebon: When would this be nil?

cmd/daemon/main.go
```diff
@@ -32,6 +34,9 @@ func init() {
 }
 
 func main() {
+	var pollInterval time.Duration
+	flag.DurationVar(&pollInterval, "poll-interval", 5*time.Minute, "Interval for polling bootc status as a fallback to fsnotify")
```
jlebon: Perhaps `bootc-pool-interval`?

internal/daemon/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+)
+
+const (
+	// ostree backend
+	DefaultPrimaryPath = "/proc/1/root/ostree/bootc"
+	// composefs backend
+	DefaultFallbackPath = "/proc/1/root/sysroot/state/deploy"
+)
+
+type StatusWatcher struct {
```
jlebon, line 27: So the way I was thinking about this was more like if it were an "informer cache" for bootc status.

So here we would maintain the deserialized status, and the daemon whenever it needed it, would query the status watcher and that either returns the in-memory object (if fsnotify is known healthy), or run `bootc status --json`.

The polling fallback still makes sense I guess to keep the BootcNode status up to date even if the daemon itself doesn't need to do any work. But ideally we only start polling once fsnotify fails?

WDYT?

internal/daemon/reconciler.go
alicefr: it was added for the unit tests. But we can initialize it there as well. Good catch

alicefr:
Implemented the caching mechanism suggested by https://github.com/bootc-dev/bootc-operator/pull/59#discussion_r3469347796 mostly included in commit: 48032ff238234f66fcc8c9c0066f8041911b79b9

internal/daemon/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+		case <-ticker.C:
+			log.V(1).Info("Polling bootc status")
+			w.sendEvent()
+		}
+	}
+}
+
+func (w *StatusWatcher) setupFsnotify(log logr.Logger, watchPath string) *fsnotify.Watcher {
+	if watchPath == "" {
+		log.Info("No bootc status path found, using polling only")
```
jlebon: Maybe "No bootc paths to watch found, using polling only"?

internal/daemon/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+	return ""
+}
+
+func (w *StatusWatcher) sendEvent() {
+	ev := event.GenericEvent{
+		Object: &bootcv1alpha1.BootcNode{
+			ObjectMeta: metav1.ObjectMeta{Name: w.NodeName},
+		},
+	}
+	select {
```
jlebon, line 202: My channel-foo is a bit rusty. What's the advantage of `select` here vs just doing `w.Events <- ev`?

internal/daemon/watcher.go
```diff
@@ -0,0 +1,135 @@
...
+	if w.Ready != nil {
+		close(w.Ready)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case ev := <-evCh:
+			// bootc updates modify directory mtime, which inotify reports as IN_ATTRIB (Chmod).
```
jlebon: Is this true for both backends?

internal/daemon/watcher.go
```diff
@@ -83,13 +136,25 @@ func (w *StatusWatcher) Start(ctx context.Context) error {
...
 }
 
+// refreshAndNotify reads bootc status, updates the cache, and sends
+// an event to trigger reconciliation.
+func (w *StatusWatcher) refreshAndNotify(ctx context.Context, log logr.Logger) {
+	if _, err := w.refresh(ctx); err != nil {
+		log.Error(err, "Failed to refresh bootc status")
+		return
+	}
+	w.sendEvent()
```
jlebon, line 161: Shouldn't we only send an event if the status actually changed?

internal/daemon/watcher.go
```diff
@@ -0,0 +1,200 @@
...
+	bootcv1alpha1 "github.com/bootc-dev/bootc-operator/api/v1alpha1"
+	"github.com/bootc-dev/bootc-operator/internal/bootc"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+)
+
+const (
+	// ostree backend
+	DefaultPrimaryPath = "/proc/1/root/ostree/bootc"
+	// composefs backend
+	DefaultFallbackPath = "/proc/1/root/sysroot/state/deploy"
```
jlebon: Primary and Fallback wording seems odd here. We should consider them on equal footing.

I'd just use Ostree and Composefs

cmd/daemon/main.go
```diff
@@ -62,17 +67,33 @@ func main() {
...
+
+	watcher := &daemon.StatusWatcher{
+		PollInterval: pollInterval,
+		PrimaryPath:  daemon.DefaultPrimaryPath,
+		FallbackPath: daemon.DefaultFallbackPath,
+		Events:       make(chan event.GenericEvent, 1),
+		NodeName:     nodeName,
+		Executor:     executor,
+	}
+	if err := mgr.Add(watcher); err != nil {
```
jlebon, line 80: Wait, how does this work? Is it the `Start()` definition that allows this to work? Might be worth adding a docstring on that `Start()` to say it implements whatever the interface name is.

internal/daemon/suite_test.go
```diff
@@ -65,12 +64,19 @@ func TestMain(m *testing.M) {
...
 
-	if err := (&BootcNodeReconciler{
+	fake := &fakeExecutor{}
+	testReconciler = &BootcNodeReconciler{
 		Client:   mgr.GetClient(),
 		Scheme:   mgr.GetScheme(),
 		NodeName: testNodeName,
 		Executor: fake,
-	}).SetupWithManager(mgr); err != nil {
+		StatusWatcher: &StatusWatcher{
```
jlebon, line 73: Don't we need to `mgr.Add()` it here too?

jlebon (commented):
Mostly nits, though one comment about change detection I think is important.

Otherwise looks sane to me.

internal/daemon/watcher.go
alicefr: The select with default makes the send non-blocking. If the channel is full, the event is just dropped instead of blocking the watcher goroutine. In this case, we might just drop a duplicate

internal/daemon/suite_test.go
alicefr: Yes, you are right here we never start the cache

cmd/daemon/main.go
alicefr: The StatusWatcher.Start satisfies the `manager.Runnable` interface from controller-runtime

internal/daemon/watcher.go
alicefr: For bootc upgrade and switch , a CREATE event is fired:
```
inotifywait  /proc/1/root/sysroot/state/deploy
Setting up watches.
Watches established.
/proc/1/root/sysroot/state/deploy/ CREATE,ISDIR 7e3177816f2cb7075289d97a3952e15b79886686eb85e25647c8410ef0b64722729e853affb71134d7aed3cf1357e1dba7fee2473ff8bb32168e138a6cac758f
```
I will add this to the code

internal/daemon/watcher.go
```diff
@@ -0,0 +1,208 @@
...
+	w.mu.RUnlock()
+	if started && s != nil {
+		return s, nil
+	}
+	s, _, err := w.refresh(ctx)
+	return s, err
+}
+
+// refresh reads bootc status from the host and updates the cache.
+// Returns the new status or an error.
```
jlebon: Docstring needs update.

internal/daemon/watcher.go
```diff
@@ -0,0 +1,208 @@
...
+	data, err := w.Executor.Status(ctx)
+	if err != nil {
+		return nil, false, err
+	}
+	s, err := bootc.ParseStatus(data)
+	if err != nil {
+		return nil, false, err
+	}
+	w.mu.Lock()
+	changed := !bytes.Equal(data, w.lastRaw)
```
jlebon: But why keep lastRaw and not just compare the `bootc.Status` objects themselves?

jlebon (dismissed):
Minor nits, but LGTM as is. CI is failing though.

internal/daemon/watcher.go
alicefr: It was for avoiding DeepEqual, but it is more readable with DeepEqual instead of raw json comparison.

jlebon (approved):

Issue #61, opened 2026-06-11
Support I/O scheduling limits for staging operations


jlebon:
See the `IO*` directives in systemd, notably e.g. `IOSchedulingClass=idle` is a common approach for "background operations". Though I can imagine also wanting something a bit more aggressive if it's on-request updates (e.g. within a time window). Probably needs to be customization at the pool level.

Calls to `bootc switch` would be passed through `systemd-run` instead so we can apply these.

alicefr:
should be the IO ops be controlled by the cluster-admin or the user? Or do we believe that the cluster-admin is always the user in our case? I'm asking because the answer determines if the IO operations should be expose in the bootc pool or rather should be something configurable for the operator and not visible to the users.

jlebon:
> should be the IO ops be controlled by the cluster-admin or the user? Or do we believe that the cluster-admin is always the user in our case? I'm asking because the answer determines if the IO operations should be expose in the bootc pool or rather should be something configurable for the operator and not visible to the users.

Anyone who has write access to the pool has full control over the node image. I/O knobs are just dynamic configs that could live in the image itself.

(But you do make me thing of a related idea, which would be to have a separate role that only allows "safer" operations e.g. upgrading nodes (for tagged pullspecs), starting/pausing/stopping rollouts, etc. Though offhand that should probably live at a higher abstraction level than this project...)

jlebon:
Related, but I don't think quite what we want: https://github.com/bootc-dev/bootc/issues/1375

PR #67, opened 2026-06-20
Various follow-ups to milestone 4c

jlebon:
These are most of the things that came up as follow-ups as part of https://github.com/bootc-dev/bootc-operator/pull/50.

### DIFF
--- a/internal/bootc/executor.go
+++ b/internal/bootc/executor.go
@@ -7,7 +7,9 @@ import (
 	"fmt"
 	"os/exec"
 	"strings"
+	"time"
 
+	"github.com/go-logr/logr"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
@@ -31,7 +33,7 @@ func NewHostExecutor() *HostExecutor {
 func (e *HostExecutor) nsenterCmd(ctx context.Context, args ...string) *exec.Cmd {
 	base := []string{
 		"--target", "1",
-		"--mount", "--pid",
+		"--mount", "--pid", "--cgroup",
 		"--setuid", "0", "--setgid", "0",
 		"--env", "--",
 	}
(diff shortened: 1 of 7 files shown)


internal/bootc/executor.go
```diff
@@ -47,19 +49,92 @@ func (e *HostExecutor) Status(ctx context.Context) ([]byte, error) {
...
+// stageUnitName is the systemd transient unit name used for bootc switch.
+const stageUnitName = "bootc-operator-switch.service"
+
 func (e *HostExecutor) Stage(ctx context.Context, image string) error {
 	log := logf.FromContext(ctx)
 
-	// TODO: use --download-only once available (https://github.com/bootc-dev/bootc/issues/2137)
-	cmd := e.nsenterCmd(ctx, "bootc", "switch", image)
+	// Stop any stale unit from a previous daemon incarnation.
+	e.stopStageUnit()
```
alicefr, line 65: What if there is still an in-progress staging operation functionally working? In this case, we stop it and recreate it right?
Don't we need simply need to re-attach to it or adopting it?
AFAIU, we wanted to run the staging operation with systemd if the daemon was restarted, e.g for upgrades, and not to stop the operation.

alicefr:
The discussion about launching the switch operation via systemd is [here](https://github.com/bootc-dev/bootc-operator/pull/50#discussion_r3450126123). Maybe we should continue it in this PR or create na ad-hoc issue

alicefr:
From the offline discussion, we decided to relay on systemd-run and the memory consumed by the `bootc switch` is part of the host OS consumption.

internal/bootc/executor.go
jlebon: I started looking at what the "adopt" path would look like when the daemon restarts, and it works but it's non-trivial. Hesitating paying for that complexity for an event that should be somewhat rare, when the current behaviour will still eventually give convergence (even if less efficiently than it could be).

This patch still at least buys us decoupling from bootc memory requirements and e.g. eventually doing https://github.com/bootc-dev/bootc-operator/issues/61, etc... which I see as more pressing than surviving restarts.

jlebon:
I was thinking about the idea we discussed of not needing a goroutine at all and thinking of the systemd-run as our goroutine (+ adoption, etc...), but an obvious thing I missed there is that we still need to re-reconcile once the switch is done. And the cleanest way to do that is still through a goroutine + GenericEvent rather than e.g. periodic requeues, etc.

alicefr (dismissed):

internal/bootc/executor.go
alicefr: ok, we can postpone this if we see that during the upgrade this is an issue.  Can we maybe add a comment like `If the daemon restarts the on-going switch operations are also considered stale and restarted`?

alicefr:
> I was thinking about the idea we discussed of not needing a goroutine at all and thinking of the systemd-run as our goroutine (+ adoption, etc...), but an obvious thing I missed there is that we still need to re-reconcile once the switch is done. And the cleanest way to do that is still through a goroutine + GenericEvent rather than e.g. periodic requeues, etc.

Thinking if we really need to reconcile once we merge https://github.com/bootc-dev/bootc-operator/pull/59 . In theory, bootc switch should fire fsnotify and we should reconcile there. Maybe we can actually avoid the go routine. The system-run should actually behave in the same way as an externally triggered switch.

jlebon:
> Thinking if we really need to reconcile once we merge #59 . In theory, bootc switch should fire fsnotify and we should reconcile there. Maybe we can actually avoid the go routine. The system-run should actually behave in the same way as an externally triggered switch.

Heh, I thought about that too, but the problem is handling the error path. If the switch failed, we still want to be woken up.

internal/bootc/executor.go
jlebon: Added a comment!

alicefr (approved):

Issue #14, opened 2026-05-07
Also manage the controller and daemonset manifests


jlebon:
See https://github.com/jlebon/bootc-operator/pull/12#discussion_r3192607381. It's a common operator pattern. E.g. for OCP, the CVO just updates the manifest for the MCO operator itself, and the latter is what updates the manifests for the MCD, MCS, MCC, etc...

alicefr:
@jlebon we haven't decided if the daemon set is deployed separately or by the operator. I would be in favor that the operator does it. It would also help for the upgrades (of the operator and the daemon set). The new version of the operator needs to take care to update the image of the daemon set, (and maybe eventually the daemon set definition as well)

jlebon:
We discussed this a bit today. IIRC, Alice mentioned that KubeVirt directly builds the resource objects and creates them that way. OTOH, the MCO has actual YAML manifests which get baked into the operator image and then read and applied, though there are concerns around memory due to the deserialization and serialization.

alicefr:
I think one of the advantages is when we update the operator. It can controls when the daemon is updated. If you update the image of the daemon set the daemon pods will be restarted. We might not want to do that if there there are nodes that are in staging/booting phase

jlebon:
We discussed this more today. We reached consensus on matching the MCO, i.e.:
- shove all the manifests in the container image
- add a new `cmd/operator` binary that syncs its baked manifests
- add a deployment manifest for the operator

Then, updating is just updating the operator deployment manifest.

This also ties into https://github.com/jlebon/bootc-operator/issues/38. Once we have published images, the manifests can reference a specific pullspec.

cheesesashimi:
Another thing that this component should be tasked with is ensuring that all garbage is collected if / when the operator deployment is deleted. See https://github.com/bootc-dev/bootc-operator/issues/160 about handling deletions gracefully.

PR #75, opened 2026-06-24
Add tag-based image resolution to the controller  (milestone 5a)

alicefr:
Pools can now reference node images by tag (e.g. registry.example.com/node:latest) instead of requiring a pinned digest. The controller periodically resolves tags to digests using go-containerregistry and updates the pool status with the resolved digest.

The resolution interval is configurable via --tag-resolution-interval (default 5m). Resolution failures are surfaced as a PoolRegistryError degraded condition. An `--allow-insecure-registry` flag enables fallback to HTTP for registries that do not serve TLS.


alicefr:
@jlebon I opened this draft PR to discuss with you a couple of design points, and I'd love to hear you opinion.

First, how to we expose the time when the tag resolution will take place. In this version, the users see when the next resolution will take place in the bootc pool status in the field `nextTagResolutionTime` (commit: https://github.com/bootc-dev/bootc-operator/pull/75/changes/fd9c1a2894241ae7f7ae8ff3b80e232e47ee3613).

Secondly, the polling time interval is configurable on the operator side by a flag. IMO, this is something the cluster admin needs to configure. But an alternative could be that it is the user who decides. In this case, we could expose the polling time under the pool spec.

Additionally, I added the supported for untrusted registry if the cluster admin configures the operator with `--allow-insecure-registry`. First, the controller tries with https, if it fails and the untrusted registries are enabled, then it tries with http.

Please, let me know what you think

alicefr:
We also need to move away from alicefr/bink repository and include the fix https://github.com/bootc-dev/bink/pull/71

jlebon:
> We also need to move away from alicefr/bink repository and include the fix [bootc-dev/bink#71](https://github.com/bootc-dev/bink/pull/71)

:arrow_right: https://github.com/bootc-dev/bootc-operator/pull/76

internal/controller/bootcnodepool_controller.go
```diff
@@ -239,12 +243,22 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	if err != nil {
 		if isInvalidSpecError(err) {
 			return r.setInvalidSpecCondition(ctx, &pool, err)
 		}
 		return ctrl.Result{}, fmt.Errorf("resolving target digest: %w", err)
 	}
+	if pool.Status.TargetDigest == "" {
+		// First tag resolution failed — nothing to roll out yet.
+		if !reflect.DeepEqual(pool.Status, *statusOrig) {
+			if err := r.Status().Update(ctx, &pool); err != nil {
```
alicefr, line 262: An alternative is to update here is to move it at the end. I had this version but I had to encapsulate the entire logic in a long if statement and I preferred to have an early return here with the update

jlebon:
> First, how to we expose the time when the tag resolution will take place. In this version, the users see when the next resolution will take place in the bootc pool status in the field `nextTagResolutionTime` (commit: [fd9c1a2](https://github.com/bootc-dev/bootc-operator/commit/fd9c1a2894241ae7f7ae8ff3b80e232e47ee3613)).

Seems fine to me.

> Secondly, the polling time interval is configurable on the operator side by a flag. IMO, this is something the cluster admin needs to configure. But an alternative could be that it is the user who decides. In this case, we could expose the polling time under the pool spec.

I touched on this a bit in another discussion, but yeah I think there _might_ be a story for a distinction between a cluster admin persona and a "bootc-operator admin" persona. I'm thinking of a managed K8s or HyperShift-like setup, where you may have a user own nodes and schedule their workloads on their nodes but not have full cluster access. But AFAIK actual HyperShift hosted clusters don't actually have edit access to their nodes. The nodes are managed via NodePools, which reside at the management cluster level. And in the case of a managed K8s cluster like EKS, you do actually have full access to the cluster.

So, there might be distinction to draw there in the future, but it's not clear to me yet how prevalent it is in the real world. So I would for now assume that cluster admin = bootc-operator admin. To circle back to the topic at hand though, I think let's leave it as a flag option for now.

> Additionally, I added the supported for untrusted registry if the cluster admin configures the operator with `--allow-insecure-registry`. First, the controller tries with https, if it fails and the untrusted registries are enabled, then it tries with http.

Seems fine to me! (I assume you had to do this for testing?)

Aside but I think at some point, we should also support configuring all operator knobs via a configmap instead. That makes it easier to separate manifest definitions from configs (related to https://github.com/bootc-dev/bootc-operator/issues/14). The configmap keys and CLI flags should just match clearly to each other.

internal/registry/resolver.go
```diff
@@ -0,0 +1,29 @@
...
+// TagResolver resolves a container image reference to a digest.
+type TagResolver interface {
+	Resolve(ctx context.Context, ref string) (string, error)
+}
+
+// GGCRResolver resolves image tags to digests using go-containerregistry.
+type GGCRResolver struct{}
+
+func (r *GGCRResolver) Resolve(ctx context.Context, ref string) (string, error) {
+	parsed, err := name.ParseReference(ref)
```
jlebon, line 22: Is it worth sanity-checking here that it's tag-based? And short-circuit if it's digest-based (and return the digest).

internal/registry/resolver_test.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func TestResolveInvalidRef(t *testing.T) {
+	g := NewWithT(t)
+
+	resolver := &GGCRResolver{}
+	_, err := resolver.Resolve(context.Background(), "")
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("parsing reference"))
+}
```
jlebon: Personally, this kind of test is not worth keeping IMO.

api/v1alpha1/bootcnodepool_types.go
```diff
@@ -45,6 +45,10 @@ const (
 	// digest ref, or a malformed nodeSelector).
 	PoolInvalidSpec string = "InvalidSpec"
 
+	// PoolRegistryError means the controller failed to contact the
+	// container registry to resolve a tag to a digest.
+	PoolRegistryError string = "RegistryError"
```
jlebon: Should we generalize this to e.g. `PoolTagResolutionError`? I.e. it's anything to do with tag resolution, which yes, the main fallible part is the registry contacting.

jlebon (commented):
Partial review

internal/registry/resolver.go
alicefr: It is called already with tag only [here](https://github.com/alicefr/bootc-operator/blob/milestone-5a/internal/controller/bootcnodepool_controller.go#L319). But the check is cheap enough worth adding it

internal/registry/resolver.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+)
+
+// TagResolver resolves a container image reference to a digest.
+type TagResolver interface {
```
jlebon: Why an interface for this?

internal/controller/bootcnodepool_controller.go
```diff
@@ -239,12 +243,22 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+				return ctrl.Result{}, fmt.Errorf("updating pool status: %w", err)
+			}
+		}
```
jlebon, line 265: Would be nice to dedupe here with the other one in this function. A completely separate function feels like overkill, and `defer` wouldn't be quite right, but perhaps just defining a closure function after line 243?

internal/registry/resolver.go
```diff
@@ -0,0 +1,29 @@
+package registry
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
```
jlebon, line 11: Did you check if there's a containers/image package we could use here? Would prefer that if it's available (_and_ it doesn't pull in tons of deps).

internal/controller/bootcnodepool_controller.go
```diff
@@ -274,24 +288,56 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		log.Error(err, "Failed to resolve tag", "ref", pool.Spec.Image.Ref)
+		apimeta.SetStatusCondition(&pool.Status.Conditions, metav1.Condition{
+			Type:    bootcv1alpha1.PoolDegraded,
+			Status:  metav1.ConditionTrue,
+			Reason:  bootcv1alpha1.PoolRegistryError,
+			Message: err.Error(),
+		})
+		next := metav1.NewTime(now.Add(r.TagResolutionInterval))
+		pool.Status.NextTagResolutionTime = &next
+		return ctrl.Result{RequeueAfter: r.TagResolutionInterval}, nil
```
jlebon: Should dedupe with the happy path flow below.

E.g. just put the TargetDigest update in the else block.

test/e2e/bootcnode_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
```
jlebon: Ahh gotcha, you're also using it for updating the tag. OK.

test/e2e/bootcnode_test.go
```diff
@@ -188,3 +190,172 @@ func TestUpdateReboot(t *testing.T) {
...
+
+const (
+	controllerNamespace  = "bootc-operator"
+	controllerDeployment = "bootc-operator-controller-manager"
+)
+
+// setTagResolutionInterval patches the controller deployment to use
+// the given interval and waits for the rollout to complete. The
+// original args are restored in t.Cleanup.
+func setTagResolutionInterval(t *testing.T, interval string) {
```
jlebon: Hmm, why not just change the `deploy-bink` target directly? Even for manual testing, a shorter interval is desirable.

jlebon (commented):
Initial pass but looks sane overall!

jlebon:
Some of the later commits could be squashed.

test/e2e/bootcnode_test.go
alicefr: I thought it will be better to set in the tests because if we run the integration tests against another cluster (like Openshift) it won't be apply it if we set it only in `deploy-bink`

internal/registry/resolver.go
alicefr: I generally prefer to abstract the structs with an interface. Additionally, it is also allows us to mock the interfaces in the tests eventually.

In KubeVirt we defined some [coding conventions](https://github.com/kubevirt/kubevirt/blob/main/docs/coding-conventions.md#short-coding-conventions-tldr), to use interfaces over the struct. Having said that, I should probably move the `TagResolver` inteface in the bootc controller

internal/registry/resolver.go
alicefr: No I used what was suggested in the implementation plan, but we can evaluate alternative packages

alicefr: AI is suggesting that containers/image pulls more dependencies (haven't verified yet) . Alternatively, we could implement our own package that talks directly to the registry

test/e2e/bootcnode_test.go
alicefr: Yeah, but if you want less dependencies we can implement our own package. Even if I would prefer to do it in a follow-up PR to keep this focusing on the digest resolution

internal/controller/bootcnodepool_controller.go
alicefr: Do you like go closures? I know it is a duplication, but I find a bit more readable

internal/registry/resolver.go
alicefr: I quickly created a [branch](https://github.com/alicefr/bootc-operator/commit/f9cae04002c1b7028aec473d76b795b79406a856) that replaces the google registry with `github.com/containers/image/`. The google registry introduces less dependecies:
```bash 
# Version with the google registry
$ git diff main..milestone-5a -- go.sum | grep '^+' | grep -v '^+++' | wc -l
26
$ git diff main..milestone-5a-registry -- go.sum | grep '^+' | grep -v '^+++' | wc -l
84
```

alicefr:
I solved some of the comments I resolved. Squashed some of the latest commits. 

But I have added the commit 2eef5ec05b5e6b91242a1ed4ace1b8ce8f28a1e4 which introduces the helper functions and resolve in a more elegant way the patching of the option in the controller. Previously, it was completely replacing all the flags for the controller. Now, it replaces existing flags with the flags that the tests needs. In this way, if the there are unrelated flags, they are preserved and tested

internal/registry/resolver.go
jlebon: WFM!

Yeah, a fake resolver for unit tests is where I thought this was going, so did a double-take when I didn't see one.

internal/registry/resolver.go
jlebon: Ack, thanks for checking.

I don't think we should implement our own package.

One thing to keep in mind is pull secrets once we support that we need to make sure go-containerregistry supports it (I would be surprised if it didn't).

test/e2e/bootcnode_test.go
jlebon: Our e2e tests already hard require bink though so I'm OK scoping it to `deploy-bink` for now.

For longer-term and for end-users, I think we should do something like https://github.com/bootc-dev/bootc-operator/issues/92.

internal/controller/bootcnodepool_controller.go
```diff
@@ -274,24 +292,53 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	ref, err := parseImageRef(pool.Spec.Image.Ref)
 	if err != nil {
-		return newInvalidSpecError(fmt.Sprintf("invalid image ref %q: %v", pool.Spec.Image.Ref, err))
+		return ctrl.Result{}, newInvalidSpecError(fmt.Sprintf("invalid image ref %q: %v", pool.Spec.Image.Ref, err))
 	}
+
 	digested, ok := ref.(reference.Digested)
-	if !ok {
-		return newInvalidSpecError(fmt.Sprintf("image ref %q has no digest (tag resolution not yet supported)", pool.Spec.Image.Ref))
+	if ok {
```
jlebon, line 311: Shouldn't we set to nil the `NextTagResolutionTime` here? This is the case where you go from a tagged ref to a digested one.

internal/controller/bootcnodepool_controller.go
```diff
@@ -239,12 +247,22 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+	resolveResult, err := r.resolveTargetDigest(ctx, &pool)
+	if err != nil {
 		if isInvalidSpecError(err) {
 			return r.setInvalidSpecCondition(ctx, &pool, err)
 		}
 		return ctrl.Result{}, fmt.Errorf("resolving target digest: %w", err)
 	}
+	if pool.Status.TargetDigest == "" {
+		// First tag resolution failed — nothing to roll out yet.
+		if !reflect.DeepEqual(pool.Status, *statusOrig) {
```
jlebon, line 261: I still feel iffy about duplicating this. The concern is more that it's not obvious that this is just "early happy path return" logic here and not specific to this return path. And that it could be missed if we ever adapt/extend what early happy path return means.

So, I'm not a fan of closures, but still would prefer it.

Not a blocker.

internal/registry/resolver_test.go
```diff
@@ -0,0 +1,65 @@
...
+	srv := httptest.NewServer(registry.New())
+	defer srv.Close()
+
+	// Build a minimal OCI image (empty base + one random layer) to push.
+	layer, err := random.Layer(256, types.OCILayer)
+	g.Expect(err).NotTo(HaveOccurred())
+	img, err := mutate.AppendLayers(empty.Image, layer)
+	g.Expect(err).NotTo(HaveOccurred())
+
+	ref, err := name.ParseReference(fmt.Sprintf("%s/test/image:latest", srv.Listener.Addr().String()))
```
jlebon: Let's also test without `:latest`.

internal/controller/bootcnodepool_controller.go
alicefr: Yes, good catch. Also a good test to add

internal/controller/bootcnodepool_controller.go
alicefr: Added the closure

internal/registry/resolver_test.go
alicefr: Add a table in the test with 2 different tags

test/e2e/bootcnode_test.go
alicefr: Right, I added the deployment of the controller with the flags in the deploy (see a70e465601fac1ac67dd7de66419b9fb04b847d4)

internal/registry/resolver_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+	"github.com/google/go-containerregistry/pkg/v1/random"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	. "github.com/onsi/gomega"
+)
+
+func TestResolveValidTag(t *testing.T) {
+	tags := []string{"latest", "v1.0"}
```
jlebon: Sorry, what I mean here is to test without any tag at all (i.e. implied `:latest`).

internal/registry/resolver_test.go
```diff
@@ -0,0 +1,69 @@
...
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(remote.Write(ref, img)).To(Succeed())
+
+			want, err := img.Digest()
+			g.Expect(err).NotTo(HaveOccurred())
+
+			resolver := &GGCRResolver{}
+			got, err := resolver.Resolve(context.Background(), ref.String())
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(got).To(Equal(want.String()))
```
jlebon: Does all this have to happen on every iteration? Feels like some of the setup stuff could be lifted out of the loop.

internal/controller/bootcnodepool_controller.go
```diff
@@ -239,13 +247,30 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+			if err := r.Status().Update(ctx, &pool); err != nil {
+				return ctrl.Result{}, fmt.Errorf("updating pool status: %w", err)
+			}
+		}
+
+		return resolveResult, nil
+	}
+
+	if pool.Status.TargetDigest == "" {
```
jlebon, line 270: The comment you had here before was good (re. first tag resolution failure).

internal/controller/bootcnodepool_controller.go
```diff
@@ -267,31 +292,56 @@ func (r *BootcNodePoolReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-		return newInvalidSpecError(fmt.Sprintf("invalid image ref %q: %v", pool.Spec.Image.Ref, err))
+		return ctrl.Result{}, newInvalidSpecError(fmt.Sprintf("invalid image ref %q: %v", pool.Spec.Image.Ref, err))
 	}
+
 	digested, ok := ref.(reference.Digested)
-	if !ok {
-		return newInvalidSpecError(fmt.Sprintf("image ref %q has no digest (tag resolution not yet supported)", pool.Spec.Image.Ref))
+	if ok {
+		pool.Status.TargetDigest = digested.Digest().String()
+		// Rest the NextTagResolutionTime in case we pass from a tag referenced image to a digested one.
```
jlebon: Reset

internal/controller/bootcnodepool_controller_test.go
```diff
@@ -3,12 +3,31 @@
...
+	pool := testutil.NewPool("tag-to-digest", testImageDigestRefA)
+	next := metav1.NewTime(time.Now().Add(time.Hour))
+	pool.Status.NextTagResolutionTime = &next
+
+	r := &BootcNodePoolReconciler{}
+	_, err := r.resolveTargetDigest(context.Background(), pool)
+	g.Expect(err).NotTo(HaveOccurred())
+	g.Expect(pool.Status.TargetDigest).To(Equal(testDigestA))
+	g.Expect(pool.Status.NextTagResolutionTime).To(BeNil())
+}
```
jlebon: Minor: this test is too artificial IMO. I would just drop it, but fine to keep too.

internal/registry/resolver.go
```diff
@@ -20,6 +24,13 @@ func (r *GGCRResolver) Resolve(ctx context.Context, ref string) (string, error)
 		return d.DigestStr(), nil
 	}
 	desc, err := remote.Get(parsed, remote.WithContext(ctx))
+	if err != nil && r.AllowInsecure {
```
jlebon: Minor: is there a specific typed error we could check for here instead of a blanket retry?

jlebon (dismissed):
LGTM overall. Mostly minor nits. Fine to merge as is but I guess you have to rebase anyway for conflicts. :)

Also, let's checkmark the completed milestone steps in the implementation plan.

Nice work!

jlebon (approved):
🎉

PR #89, opened 2026-07-01
Run the e2e test on multiple k8s versions

### LINKED ISSUE #48: Document and test minimum k8s version supported
Probably should just support all active K8s versions. We're not really using any fancy new APIs here... (yet)

Add a CI check for the oldest supported version and document it.

alicefr:
Test the bootc-operator on all the active kuberentes versions. The version are dynamically fetched from https://endoflife.date/api/kubernetes.json

Fixes: #48

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -17,6 +17,18 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  supported-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - name: Get supported Kubernetes versions
+        id: set-matrix
+        run: |
+          MATRIX=$(curl -s https://endoflife.date/api/kubernetes.json | \
+            jq -c '[.[] | select(.eol > (now | strftime("%Y-%m-%d"))) | .cycle][:3]')
+          echo "matrix=${MATRIX}" >> "$GITHUB_OUTPUT"
+
   unit:
     runs-on: ubuntu-latest
     permissions:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -158,12 +180,12 @@ jobs:
...
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
-          name: e2e-logs
+          name: e2e-logs-${{ matrix.kube-minor }}
           path: _output/logs/
           if-no-files-found: ignore
 
       - name: Push to GHCR
-        if: github.event_name == 'push'
+        if: github.event_name == 'push' && matrix.kube-minor == steps.defaults.outputs.kube-minor
```
jlebon: Hmm, I understand why we need this (so that it only happens once), but that's not obvious from this. I wonder if there's a more GHA native way to do this that still avoids having to rebuild the image. Otherwise, let's at least add a comment.

jlebon (commented):
Ideally we'd have a single buildimg in a separate job, push it to ghcr, and then the matrix test, and then another job for pushing. But pushing to ghcr is privileged and I don't want to add those privs (I think Zizmor would also complain). We could `podman save` and `podman load` I guess? It's hacky but the image is not that large either.

alicefr:
> Ideally we'd have a single buildimg in a separate job, push it to ghcr, and then the matrix test, and then another job for pushing. But pushing to ghcr is privileged and I don't want to add those privs (I think Zizmor would also complain). We could `podman save` and `podman load` I guess? It's hacky but the image is not that large either.

I also don't like to push to ghcr at every run of the CI and untested images.

Makefile
```diff
@@ -21,6 +22,9 @@ all: build
 
 ##@ General
 
+print-var-%:
+	@echo $($*)
+
```
jlebon, line 27: Minor: don't really need this anymore. But fine to keep too.

jlebon (approved):
Thank you, this is great!

I had to change the ruleset for `main` to no longer require `e2e` and instead require `e2e (1.34)`, `e2e (1.35)`, and `e2e (1.36)`.

That's fine for now, but perhaps we should add a simple e.g. `e2e-collect` which just `needs` all the matrix branches. And then the ruleset can just require that one.

Not sure, perhaps there's a cleaner way to do this.

PR #91, opened 2026-07-02
UpdateReboot with Composefs disk image

### LINKED ISSUE #74: Also test composefs node images in CI
Bink builds composefs disk images now that we should be able to use in CI.

Probably shouldn't X 2 every test, but instead pick some tests that meaningfully exercise bootc stuff and opt them into multi-backend testing.

HarshwardhanPatil07:
UpdateReboot with Composefs disk image

Closes #74

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,7 +8,7 @@ on:
     branches: [main]
 
 env:
-  BINK_COMMIT: 3ddb9da3b6e33f5dadd48abdb15ed550bbfe4f31
+  BINK_COMMIT: 61899a68d69b0de1114e165d24364bf5b8b6da2e
 
 permissions: {}
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -172,6 +172,11 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y podman
 
+      # Ubuntu runners ship a v1 (INI) registries.conf that Podman >= 5
+      # rejects. Overwrite with a minimal v2 (TOML) config.
+      - name: Write v2 registries.conf
+        run: printf 'unqualified-search-registries = ["docker.io"]\n' | sudo tee /etc/containers/registries.conf
+
       - name: Start podman socket
         run: systemctl --user start podman.socket
 
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ KUBECONFIG_BINK ?= ./kubeconfig-$(BINK_CLUSTER_NAME)
 ARTIFACTS ?= $(abspath _output/logs)
 DEFAULT_KUBE_MINOR ?= 1.35
 BINK_NODE_DISK_IMAGE ?= ghcr.io/bootc-dev/bink/node:v$(DEFAULT_KUBE_MINOR)-fedora-44-disk
+BINK_NODE_DISK_IMAGE_COMPOSEFS ?= $(BINK_NODE_DISK_IMAGE)-composefs
 BINK_LOCAL_REGISTRY_NODE_IMAGE ?= registry.cluster.local:5000/node
 # YEAR defines the year value used for substituting the YEAR placeholder in the boilerplate header.
 YEAR ?= $(shell date +%Y)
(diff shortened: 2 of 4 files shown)


test/e2e/e2eutil/env.go
```diff
@@ -140,6 +141,15 @@ func WithLabel(key, value string) NodeOption {
 	}
 }
 
+// WithNodeDiskImage overrides the VM disk image passed as --node-image
+// to bink node add. When unset, AddNode falls back to the
+// BINK_NODE_DISK_IMAGE environment variable.
```
alicefr: This isn't what the function does tough. If you aren't falling back to the env variable

Makefile
```diff
@@ -7,6 +7,7 @@ BINK_CLUSTER_NAME ?= e2e
 KUBECONFIG_BINK ?= ./kubeconfig-$(BINK_CLUSTER_NAME)
 ARTIFACTS ?= $(abspath _output/logs)
 BINK_NODE_DISK_IMAGE ?= ghcr.io/bootc-dev/bink/node:v1.35-fedora-44-disk
+BINK_NODE_DISK_IMAGE_COMPOSEFS ?=
```
alicefr: Why not already setting here the default as the  `BINK_NODE_DISK_IMAGE `. In this way you can also avoid to add the variable in the GHA action and relay on the defaults

test/e2e/bootcnode_test.go
```diff
@@ -95,12 +95,29 @@ func TestControllerMembership(t *testing.T) {
 // original image, then updates the pool to a new image and verifies the
 // full update lifecycle: staging, reboot, and idle with the new image.
 func TestUpdateReboot(t *testing.T) {
+	testUpdateReboot(t)
```
alicefr: An alternative is to create some kind of table with the 2 different image:
```suggestion
func TestUpdateReboot(t *testing.T) {
	tests := []struct {
		name   string
		envVar string
	}{
		{
			name: "ostree",
			envVar: "BINK_NODE_DISK_IMAGE"
		},
		{
			name:   "composefs",
			envVar: "BINK_NODE_DISK_IMAGE_COMPOSEFS",
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			var nodeOpts []e2eutil.NodeOption
			if tc.envVar != "" {
				img := os.Getenv(tc.envVar)
				nodeOpts = append(nodeOpts, e2eutil.WithNodeDiskImage(img))
			}
			testUpdateReboot(t, nodeOpts...)
		})
	}
}

**Comment by alicefr** [comment]
@HarshwardhanPatil07 many thanks for the PR! A couple of suggestions. Additionally, I would completely drop the second commit and set the default directly in the Makefile.

**Comment by alicefr** [comment]
@HarshwardhanPatil07 you also need to bump the bink commit in the [GHA](https://github.com/bootc-dev/bootc-operator/blob/main/.github/workflows/ci.yaml#L11) to the latest inorder to pick up the composefs fix in bink

File: test/e2e/bootcnode_test.go (hunk 3 above)
Comment (alicefr) [reply test/e2e/bootcnode_test.go]: you could also generalize this even further an create a function that can be reused by multiple tests. Something like:
```golang
var diskImages = []diskImage{                                                                                
  {name: "ostree", envVar: "BINK_NODE_DISK_IMAGE"},                                                          
  {name: "composefs", envVar: "BINK_NODE_DISK_IMAGE_COMPOSEFS"},                                             
}                                                                                                            
                                                                                                             
func forEachDiskImage(t *testing.T, fn func(t *testing.T, nodeOpts ...e2eutil.NodeOption)) {                 
  t.Helper()                                                                                                 
 for _, di := range diskImages {                                                                            
        t.Run(di.name, func(t *testing.T) {                                                                      
        img := os.Getenv(di.envVar)                                                                            
        fn(t, e2eutil.WithNodeDiskImage(img))                                                                  
     })
  }
}
``` 
and then the test becomes
```golang
func TestUpdateReboot(t *testing.T) {                                                                        
  forEachDiskImage(t, testUpdateReboot)                                                                      
}
```

HarshwardhanPatil07:
@alicefr thanks for your review. I have added the changes. Could you please take a look

test/e2e/bootcnode_test.go
```diff
@@ -95,12 +95,44 @@ func TestControllerMembership(t *testing.T) {
...
+				if img == "" {
+					t.Skipf("%s not set", tc.envVar)
+				}
+				nodeOpts = append(nodeOpts, e2eutil.WithNodeDiskImage(img))
+			}
+			testUpdateReboot(t, nodeOpts...)
+		})
+	}
+}
+
```
alicefr: can you create a reusable function has it was suggested by the comment https://github.com/bootc-dev/bootc-operator/pull/91#discussion_r3511579346 . In this way, we can reuse the same pattern in other tests as well

HarshwardhanPatil07: Yes, thank you!!

HarshwardhanPatil07:
@alicefr ready for your review! Let me know if anything needs adjusting.

HarshwardhanPatil07:
Also the tests are not triggered don't know why e2e (1.34) and those?

alicefr:
They require an approval at every time. I thought it required only once. I will ask today in the bootc meeting how they deal with it. Usually, there is a label that enables subsequent runs. 
If you are part of bootc org, then you can run automatically the tests

.github/workflows/ci.yaml
```diff
@@ -172,6 +172,9 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y podman
 
+      - name: Write v2 registries.conf
+        run: printf 'unqualified-search-registries = ["docker.io"]\n' | sudo tee /etc/containers/registries.conf
+
```
jlebon, line 179: This should be a separate commit ideally with its own rationale. And a smaller version of that rationale as a comment here.

HarshwardhanPatil07: CI test was getting failed

HarshwardhanPatil07:
I was trying something and see what happens when you delete the fork repository, This got closed liked that, sorry for experimentation. reopened it [here](https://github.com/bootc-dev/bootc-operator/pull/97). I tried pushing the branch on another fork and recover it like that but didn't worked.

HarshwardhanPatil07:
probably this is something github can add, maybe

Issue #17, opened 2026-05-08
Integrate with trusted-cluster-operator


jlebon:
See https://github.com/trusted-execution-clusters/operator. The bootc-operator handles updates, which needs tight integration with the trusted-cluster-operator.

The operator has some OpenShift integration but can be used in vanilla K8s as well.

PR #111, opened 2026-07-15
Define ROADMAP

### LINKED ISSUE #79: Add a ROADMAP.md
See https://contribute.cncf.io/projects/best-practices/community/contributor-growth/open-source-roadmaps/

alicefr:
The roadmap helps to strength the direction of the project and gives an overview of the next upcoming features. This first versions aims to collect ideas and issues raised during the past months of discussions.

Fixes: #79

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,47 @@
+# Bootc Operator Roadmap
+
+## Vision
+
+The Bootc Operator aims to be the standard Kubernetes operator for managing
+bootc-based node OS images. It bridges bootc and the Kubernetes control plane,
+enabling cluster admins to declaratively manage host OS updates, rollouts, and
+surfaces future bootc capabilities across any Kubernetes distribution. It claims
+to be distro-agnostic and bootc-native.
+
+## Roadmap Phases
+
+### Phase 1: Core Operator with minimal MVP
+
+* Complete milestone `v0.1`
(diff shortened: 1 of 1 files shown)


alicefr:
I created phases, but thinking more about it they are more categories. Phases implies kind of one phase after the other when some of them can be swapped.

jlebon:
> I created phases, but thinking more about it they are more categories. Phases implies kind of one phase after the other when some of them can be swapped.

Yeah agree. Perhaps we should just have them be categories then. I struggle whether the roadmap should be timeline-based or category-based. Timeline-based makes sense for a roadmap, but e.g. I wouldn't try to have stuff past a e.g. "phase 3" though. It's just too hard to predict where focus will need to be as requirements from the various places the operator ends up getting used become clearer (e.g. I can imagine https://github.com/bootc-dev/bootc-operator/issues/17 taking priority earlier if it calls for it).

But maybe that's for the best anyway. I'm not sure we should overcharge the doc either with so many details and issue links.

What is super important to me and which should be in there also is surfacing bootc features to the operator. A huge one is soft-reboot, which already works today at the bootc layer. But also e.g. pushing forward https://github.com/bootc-dev/bootc/issues/22. Both those things are highly relevant and desirable for K8s clusters.

OK, I started a draft locally with some of this. Interested in your thoughts though on timeline vs category.

jlebon:
Thanks for getting the ball rolling on this!

Took some of your initial wording and lists, but heavily reworked and cut it down, as well as extended with various bootc integration features: https://github.com/bootc-dev/bootc-operator/pull/118. Closing in favour of that one.

PR #112, opened 2026-07-15
controller: add precedence-aware setPoolDegraded helper

jlebon:
It's annoying right now how there's unclear precedence rules on the
PoolDegraded reasons. Rather than trying to encode this in the
imperative businesss logic, let's add a clear precedence map and helper
where the source of truth for this stuff lives.

Of course, the larger issue there is that it's even possible to have
multiple degraded reasons coexist, which suggests we should actually
split out into more condition types. Otherwise, sysadmins have to fix
one, and then wait for the next reconcile, fix the next issue, etc...

But at the same time... fully splitting them out would be quite a lot,
and I think in practice it'll be quite rare to have a pool with more
than e.g. 2 categories of bad things going on.

I also think once we start emitting Events, that will help a lot with
this visibility issue. So the API complexity of multiple condition types
might not be worth it.

### DIFF
--- a/internal/controller/bootcnodepool_controller.go
+++ b/internal/controller/bootcnodepool_controller.go
@@ -327,12 +327,7 @@ func (r *BootcNodePoolReconciler) resolveTargetDigest(ctx context.Context, pool 
 	digest, err := r.TagResolver.Resolve(ctx, pool.Spec.Image.Ref)
 	if err != nil {
 		log.Error(err, "Failed to resolve tag", "ref", pool.Spec.Image.Ref)
-		apimeta.SetStatusCondition(&pool.Status.Conditions, metav1.Condition{
-			Type:    bootcv1alpha1.PoolDegraded,
-			Status:  metav1.ConditionTrue,
-			Reason:  bootcv1alpha1.PoolTagResolutionError,
-			Message: err.Error(),
-		})
+		setPoolDegraded(pool, bootcv1alpha1.PoolTagResolutionError, err.Error())
 	} else {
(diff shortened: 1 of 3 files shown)


internal/controller/bootcnodepool_controller.go
```diff
@@ -644,18 +639,39 @@ func (r *BootcNodePoolReconciler) restoreCordonState(ctx context.Context, bn *bo
...
+	bootcv1alpha1.PoolInvalidSpec:        4,
+	bootcv1alpha1.PoolTagResolutionError: 5,
+}
+
+// setPoolDegraded sets the Degraded condition on the pool, but only if the new
+// reason has higher precedence than the current one. This prevents e.g.
+// RolloutHalted from overwriting NodeConflict. Unknown reasons get precedence
+// 0 (map zero-value), so they silently no-op. Keep poolDegradedPrecedence in
+// sync with the API constants.
+func setPoolDegraded(pool *bootcv1alpha1.BootcNodePool, reason, message string) {
```
alicefr, line 664: We don't have unit tests for such helpers functions yet, but it might be helpful to add some. The function is relatively straightforward so up to you. But it might be good to introduce it as a good habit

alicefr:
Looks good to me but you need to rebase :)

internal/controller/bootcnodepool_controller.go
jlebon: We do have a few function-level unit tests though I skipped adding one for this one because it just didn't seem worth it for the complexity.

alicefr (approved):

Issue #69, opened 2026-06-22
E2e test planning


alicefr:
Let's discuss about the test planning. Currently, we only test the happy path for the reboot 

Ideas about follow-up tests:
1. Extend the test [TestUpdateReboot](https://github.com/bootc-dev/bootc-operator/blob/main/test/e2e/bootcnode_test.go#L97C6-L97C22) by rolling back to the previous image
1. Pause the rollout and check that the resuming continue correctly
1. Try to upgrade to no-existing image, the node should go in degraded state and the rollout should be paused
1. Shutdown (manually) a node after the reboot to simulate a faulty node, after the upgrade. It should be reported as degraded  
1. Try to kill the controller/daemon during a roll-out. They should recover and the rollout should progress and terminate successfully
1. Testing the happy path with the digests (once supported)
1. Upgrade to an image protected by a pull secret (once supported) 
1. Try to switch to a completely different image e.g fedora to centos stream
1. Test upgrades for the operator

jlebon:
One general note: since each test spawns a separate worker node, which takes time, I would bundle these subtests into tests where it makes sense. E.g. error conditions/recovery tests could all be checked together in one test for example.

jlebon:
> Try to switch to a completely different image e.g fedora to centos stream

Is the more general thing we're testing here that the node degrades successfully if rebooting into an image which doesn't actually have e.g. the kubelet?

> Test upgrades for the operator

This relates to https://github.com/bootc-dev/bootc-operator/issues/14.

jlebon:
@ptalgulk01 Another useful test here is: you start a rollout for a certain image, and then change the target image during the rollout. We should still see all the nodes eventually converge to the updated target image (even if some might've already rebooted into the old one by then). This is important because it verifies that the state machines of the controller and daemon can't deadlock each other during corner cases.

ptalgulk01:
> [@ptalgulk01](https://github.com/ptalgulk01) Another useful test here is: you start a rollout for a certain image, and then change the target image during the rollout. We should still see all the nodes eventually converge to the updated target image (even if some might've already rebooted into the old one by then). This is important because it verifies that the state machines of the controller and daemon can't deadlock each other during corner cases.

thank you, I have added the test scenario here https://github.com/bootc-dev/bootc-operator/pull/115/changes#diff-f653b9b9559a51ff5f52f509882107638c85311fb6074afffc018e1b4ef9fbc0R404

cheesesashimi:
[...]

PR #115, opened 2026-07-16
e2e: add rollback, pause/resume, and non-existing image tests

ptalgulk01:
Partial progress on #69 (items 1, 2, 3):

  - **Rollback** (item 1): Extend TestUpdateReboot with phases 7-8 that roll back  to the original image and verify the node returns to Idle with the original digest
  - **Pause/Resume** (item 2): Add TestPauseResume that starts an update with paused=true, verifies the node stages but does not reboot, then resumes and verifies the update completes
  - **Non-existing image** (item 3): Add TestNonExistingImage that updates to a corrupted digest and verifies the node enters Degraded state with Reason=Error while remaining booted on the original image

 All three tests pass against a bink cluster.

TestMidRolloutImageChange (item 4) will be added in a follow-up PR.

  ## Test plan
  
  - [x] TestControllerMembership — PASS (pre-existing)
  - [x] TestUpdateReboot (with rollback) — PASS
  - [x] TestTagResolution — PASS (pre-existing)
  - [x] TestPauseResume — PASS
  - [x] TestNonExistingImage — PASS

### DIFF
--- a/test/e2e/bootcnode_test.go
+++ b/test/e2e/bootcnode_test.go
@@ -198,6 +198,35 @@ func TestUpdateReboot(t *testing.T) {
 		fmt.Sprintf("expected update-marker to exist on host, kubectl exec output: %s", string(out)))
 
 	t.Logf("Verified update-marker exists on host via daemon pod")
+
+	// Phase 7: Rollback to original image.
+	originalRef := env.NodeImageDigestedPullSpec()
+
+	modified = pool.DeepCopy()
+	modified.Spec.Image.Ref = originalRef
+	g.Expect(env.Client.Patch(ctx, modified, client.MergeFrom(pool))).To(Succeed())
+	*pool = *modified
+
+	t.Logf("Patched pool to rollback to original image %s", originalRef)
+
+	// Phase 8: Wait for Idle with the original digest — proves rollback succeeded.
+	g.Eventually(func() (bootcv1alpha1.BootcNodeStatus, error) {
+		var bn2 bootcv1alpha1.BootcNode
+		err := env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn2)
+		return bn2.Status, err
+	}).WithTimeout(5*time.Minute).Should(And(
+		HaveField("Booted", And(
+			Not(BeNil()),
+			HaveField("ImageDigest", Equal(env.NodeImageDigest())),
+		)),
+		HaveField("Conditions", ContainElement(And(
+			HaveField("Type", bootcv1alpha1.NodeIdle),
+			HaveField("Status", metav1.ConditionTrue),
+			HaveField("Reason", bootcv1alpha1.NodeReasonIdle),
+		))),
+	), "expected node to reach Idle with original image after rollback")
+
+	t.Logf("Node %q successfully rolled back to original image", nodeName)
 }
 
 // TestTagResolution creates a pool with a tag-based image ref, verifies
@@ -220,19 +249,19 @@ func TestTagResolution(t *testing.T) {
 	g.Expect(env.Client.Create(ctx, pool)).To(Succeed())
 
 	// Verify targetDigest is resolved to the original image digest.
-	g.Eventually(func(g Gomega) string {
+	g.Eventually(func() (string, error) {
 		var p bootcv1alpha1.BootcNodePool
-		g.Expect(env.Client.Get(ctx, client.ObjectKeyFromObject(pool), &p)).To(Succeed())
-		return p.Status.TargetDigest
+		err := env.Client.Get(ctx, client.ObjectKeyFromObject(pool), &p)
+		return p.Status.TargetDigest, err
 	}).WithTimeout(1 * time.Minute).Should(Equal(env.NodeImageDigest()))
 
 	t.Logf("Tag resolved to original digest %s", env.NodeImageDigest())
 
 	// Wait for node to reach Idle with the original image.
-	g.Eventually(func(g Gomega) bootcv1alpha1.BootcNodeStatus {
+	g.Eventually(func() (bootcv1alpha1.BootcNodeStatus, error) {
 		var bn bootcv1alpha1.BootcNode
-		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
-		return bn.Status
+		err := env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)
+		return bn.Status, err
 	}).WithTimeout(3 * time.Minute).Should(And(
 		HaveField("Booted", And(
 			Not(BeNil()),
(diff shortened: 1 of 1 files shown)


ptalgulk01:
cc @alicefr @jlebon can ptal?

test/e2e/bootcnode_test.go
```diff
@@ -198,86 +198,188 @@ func TestUpdateReboot(t *testing.T) {
...
+		)))
+	}).WithTimeout(5*time.Minute).Should(Succeed(), "expected node to reach Idle with original image after rollback")
+
+	t.Logf("Node %q successfully rolled back to original image", nodeName)
 }
 
-// TestTagResolution creates a pool with a tag-based image ref, verifies
-// the controller resolves the tag to a digest, then retags the image
-// and verifies re-resolution triggers a rollout.
-func TestTagResolution(t *testing.T) {
```
jlebon, line 206: Hmm wait, this seems to be dropping tag-based testing. Was that intended? Why?

ptalgulk01: thankyou for taking a look, I have addressed it.

test/e2e/bootcnode_test.go
```diff
@@ -198,6 +198,31 @@ func TestUpdateReboot(t *testing.T) {
...
+	g.Eventually(func(g Gomega) {
+		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
+		g.Expect(bn.Status.Booted).NotTo(BeNil())
+		g.Expect(bn.Status.Booted.ImageDigest).To(Equal(env.NodeImageDigest()),
+			"expected booted digest to match original image after rollback")
+		g.Expect(bn.Status.Conditions).To(ContainElement(And(
+			HaveField("Type", bootcv1alpha1.NodeIdle),
+			HaveField("Status", metav1.ConditionTrue),
+			HaveField("Reason", bootcv1alpha1.NodeReasonIdle),
+		)))
```
jlebon: Ask you agent e.g.: update the new tests to be in line with the latest `REVIEW_GOLANG.md` guidelines.

test/e2e/bootcnode_test.go
```diff
@@ -247,6 +272,13 @@ func TestTagResolution(t *testing.T) {
 	t.Logf("Node %q is Idle with original image", nodeName)
 
 	// Retag node:latest to point at the update image.
+	// Restore the original tag on cleanup so later tests are not affected.
+	t.Cleanup(func() {
+		e2eutil.RetagImage(t,
+			"localhost:5000/node@"+env.NodeImageDigest(),
+			"localhost:5000/node:latest",
+		)
+	})
```
jlebon, line 289: I would put this _after_ the `RetagImage` below.

But probably eventually it would make sense to have each test use their own tags so there's no conflict possible. Similar to pool names.

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+	g.SetDefaultEventuallyTimeout(pollTimeout)
+	g.SetDefaultEventuallyPollingInterval(pollInterval)
+
+	env := e2eutil.New(t)
+	nodeName := env.AddNode(t)
+
+	ctx := context.Background()
+
+	// Phase 1: Create pool with original image and wait for Idle.
+	pool := env.NewPool("workers", env.NodeImageDigestedPullSpec())
```
jlebon: Pool names should be unique for the test (in theory we'd like to eventually be able to parallelize tests).

Though probably to do this right eventually we should enforce it at the harness level. I.e. `NewPool() no longer takes a name but instead returns a name, similar to `AddNode`.

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+		g.Expect(bn.Status.Staged.ImageDigest).To(Equal(env.NodeImageUpdateDigest()),
+			"expected staged digest to match update image")
+		g.Expect(bn.Status.Conditions).To(ContainElement(And(
+			HaveField("Type", bootcv1alpha1.NodeIdle),
+			HaveField("Status", metav1.ConditionFalse),
+			HaveField("Reason", bootcv1alpha1.NodeReasonStaged),
+		)))
+		// Booted image should still be the original.
+		g.Expect(bn.Status.Booted.ImageDigest).To(Equal(env.NodeImageDigest()))
+	}).WithTimeout(5 * time.Minute).Should(Succeed())
```
jlebon: I would add a `g.Consistently` here which waits e.g. 10 seconds or something and verifies that the bootcnode's desiredImageState remains `Staged`.

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+			HaveField("Type", bootcv1alpha1.NodeIdle),
+			HaveField("Status", metav1.ConditionFalse),
+			HaveField("Reason", bootcv1alpha1.NodeReasonStaged),
+		)))
+		g.Expect(bn.Status.Booted.ImageDigest).To(Equal(env.NodeImageDigest()))
+	}).WithTimeout(5 * time.Minute).Should(Succeed())
+
+	t.Logf("Node %q staged update image (paused, not rebooted)", nodeName)
+
+	// Phase 4: Change the target back to the original image while still
```
jlebon: This is not very different from `TestPauseResume`. Not quite what I had in mind for https://github.com/bootc-dev/bootc-operator/issues/69#issuecomment-5060986733. Instead we do something like:
- provision two nodes (let's say A and B)
- set the new updated image
- wait for node A to be Rebooting
- immediately switch the target image to another updated image
- wait until both node A and node B eventually are running the updated image
- verify that node B did not incur an additional reboot into the first update image (e.g. check `journalctl --list-boots` perhaps?)

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+	t.Logf("Patched pool to non-existing image %s", nonExistingRef)
+
+	// Phase 3: Wait for node to enter degraded state.
+	// The daemon should fail to pull the image and report an error.
+	g.Eventually(func(g Gomega) {
+		g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
+		g.Expect(bn.Status.Conditions).To(ContainElement(And(
+			HaveField("Type", bootcv1alpha1.NodeDegraded),
+			HaveField("Status", metav1.ConditionTrue),
+			HaveField("Reason", bootcv1alpha1.NodeReasonError),
```
jlebon, line 480: Would be nice to also assert the condition message here, or at least a substring.

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+
+	t.Logf("Node %q entered degraded state as expected", nodeName)
+
+	// Phase 4: Verify the node did not stage the non-existing image.
+	g.Expect(env.Client.Get(ctx, client.ObjectKey{Name: nodeName}, &bn)).To(Succeed())
+	// Staged should either be nil or not match the non-existing digest.
+	if bn.Status.Staged != nil {
+		g.Expect(bn.Status.Staged.ImageDigest).NotTo(ContainSubstring("sha256:"),
+			"node should not have staged the non-existing image")
+	}
```
jlebon: Shouldn't it always be nil?

test/e2e/bootcnode_test.go
```diff
@@ -281,3 +313,261 @@ func TestTagResolution(t *testing.T) {
...
+
+	t.Logf("Node %q is Idle with original image", nodeName)
+
+	// Phase 2: Patch pool to update to a non-existing image.
+	// Use a digest that doesn't exist in the registry.
+	nonExistingRef := env.NodeImageDigestedPullSpec()
+	if idx := len(nonExistingRef) - 1; idx >= 0 {
+		// Modify the last character of the digest to make it invalid
+		nonExistingRef = nonExistingRef[:idx] + "0"
+	}
```
jlebon: Meh, totally fine also to just hardcode a fake digest.

jlebon (commented):
Thanks for working on this!

test/e2e/bootcnode_test.go
ptalgulk01: > immediately switch the target image to another updated image

here you mean to update the node A when it is rebooting to new image already we have patched to other new image?
And keep node B as it is

ptalgulk01:
Changes added:
 - Converted all new Eventually callbacks to func() (T, error) pattern per REVIEW_GOLANG.md
  - Moved t.Cleanup after RetagImage in TestTagResolution
  - Unique pool names per test (bnp-pause, bnp-midroll, bnp-noimg)
  - Added g.Consistently (10s) to verify node stays Staged while paused
  - Hardcoded fake digest in TestNonExistingImage
  - Added condition message substring assertion ("stage failed")
  - Replaced Staged != nil guard with g.Expect(bn.Status.Staged).To(BeNil())
  - TestMidRolloutImageChange: waiting for clarification on the 2-node design

test/e2e/bootcnode_test.go
jlebon: > > immediately switch the target image to another updated image
> 
> here you mean to update the node A when it is rebooting to new image already we have patched to other new image? And keep node B as it is

I mean changing the pool targetImage to another image. We don't mutate the nodes directly. Node A will reboot into the wrong image, which is fine and expected, but it should then realize that it's not in sync and immediately restage and reboot again. Node B should be able to avoid that reboot and directly stage the new image (abandoning the staged deployment it already had for the previous targetImage).

jlebon:
Looks sane overall. If you want you can drop the `TestMidRolloutImageChange` test from this PR and we can get it in, and then add that test in separately (though perhaps then it shouldn't be marked as closing #69).

jlebon (approved):
Thanks!

jlebon:
Interesting that the e2e (1.34) test flaked here:

```
bink registry start
Error: starting registry: creating registry container: starting container: crun: unknown version specified: OCI runtime error
Usage:
  bink registry start [flags]

Flags:
  -h, --help   help for start

Global Flags:
  -c, --cluster-name string   cluster name (default "podman")
      --config string         config file (default is $HOME/.bink/config.yaml)
      --debug                 debug output
  -v, --verbose               verbose output

starting registry: creating registry container: starting container: crun: unknown version specified: OCI runtime error
make: *** [Makefile:133: seed-node-image] Error 1
```

It passed when the test was rerun on main.

PR #121, opened 2026-07-24
Complete milestone 3b

jlebon:
As usual, see individual commits. But the main commit here is drain cancellation, and adding a test that this works as expected.

I actually thought on paper about some other drain-related error-handling but it turns out our state machine is quite good already and still converges anyway. There are some things there that are somewhat tricky to balance. E.g. let's say we've already drained a node, and _then_ the target image changes... what should we do? Currently the code just leaves the node cordoned while it stages the new commit and then eventually reboots into the new target image. That might not make sense if your connection is extremely throttled because meanwhile you're not servicing pods. Whereas probably a better approach in that case would be to uncordon on the old node, and stage in the background as usual. But anyway, at least we converge in the end. Handling this kind of "backing out" is surprisingly tricky and it didn't feel worth paying for that complexity just yet until we really hit it in the wild.

### DIFF
--- a/docs/IMPLEMENTATION_PLAN.md
+++ b/docs/IMPLEMENTATION_PLAN.md
@@ -136,7 +136,7 @@ the full controller+daemon loop can be tested end-to-end.
   nodes are labeled. This exercises the e2e harness for the first time
   with a running operator and validates it functions in a real cluster.
 
-### 3b. Digest-only rollout state machine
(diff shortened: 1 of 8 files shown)


alicefr (approved):

alicefr:
Thanks Jonathan, it looks great!

PR #127, opened 2026-08-04
makefile: Start podman socket before running bink

Johan-Liebert1:
If podman socket is not running then any and all bink commands fail, so we just start the socket ourselves before running bink commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,10 @@ e2e: ## Run e2e tests (requires: make deploy-bink). V=1 for verbose. RUN=<regex>
 		BINK_NODE_IMAGE_UPDATE_DIGEST=$$(skopeo inspect --tls-verify=false docker://localhost:5000/node:update | jq -r '.Digest') \
 		go test -timeout 20m -count=1 $(if $(V),-v) $(if $(RUN),-run $(RUN)) .
 
+.PHONY: start-socket
+start-socket:
+	systemctl start --user podman.socket
+
 ##@ Build
 
 .PHONY: build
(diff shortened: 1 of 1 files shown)


Johan-Liebert1:
ping @alicefr @jlebon

alicefr:
I'm not a big fun in starting a systemd service from a makefile. But it is true that we are missing this installation step in bink in the README. IMO, it belongs to the readme in bink not really in the bootc  operator  repository

alicefr:
Another reason not to include this is that we might want to run bink from an agent and in this case we have multiple different setups. I know that Jonathan is using nested containers while I run the remote podman in a separate container. Hence, bink might configure differently for several environment, a couple of examples are given in the bink [README](https://github.com/bootc-dev/bink/blob/main/README.md#running-via-container)

Johan-Liebert1:
That's fair I think. Maybe the real problem here is the error message that bink throws when it can't connect to Podman service

```
Error: listing containers: connecting to Podman service at unix:///run/user/1000/podman/podman.sock: unable to connect to Podman socket: Get "http://d/v6.0.0/libpod/_ping": dial unix /run/use
r/1000/podman/podman.sock: connect: connection refused
Usage:
  bink cluster list [flags]

Examples:
  # List all clusters and their status
  bink cluster list

Flags:
  -h, --help        help for list
      --name-only   Only print cluster names, one per line

Global Flags:
  -c, --cluster-name string   cluster name (default "podman")
      --config string         config file (default is $HOME/.bink/config.yaml)
      --debug                 debug output
  -v, --verbose               verbose output

listing containers: connecting to Podman service at unix:///run/user/1000/podman/podman.sock: unable to connect to Podman socket: Get "http://d/v6.0.0/libpod/_ping": dial unix /run/user/1000/
podman/podman.sock: connect: connection refused
```

Would you accept a PR in bink appending something like `Try starting podman socket using "systemd --user start podman.socket"` to the error message?

alicefr:
> Would you accept a PR in bink appending something like `Try starting podman socket using "systemd --user start podman.socket"` to the error message?

Definitely, everything that can improve the error messaging.

PR #151, opened 2026-08-26
Support soft reboot

### LINKED ISSUE #117: Support soft reboot
The operator's `BootcNodePool` spec already has a `disruption.rebootPolicy` field with `RebootOnly` (default) and `AllowSoftReboot` values, but the daemon doesn't implement soft reboot yet.

When `AllowSoftReboot` is set, the daemon should pass `--soft-reboot=auto` to `bootc switch --from-downloaded --apply`. This performs a userspace-only restart when possible (no kernel changes), avoiding a full hardware reboot. This significantly reduces disruption for updates that only change userspace content.

Upstream bootc support: `bootc switch --apply --soft-reboot=auto` is already available (see [bootc-switch.8](https://github.com/bootc-dev/bootc/blob/main/docs/src/man/bootc-switch.8.md)).

Add an e2e test that creates a pool with `AllowSoftReboot`, triggers an update, and verifies the node comes back up without a full reboot (e.g. by checking that the boot ID is preserved).

HarshwardhanPatil07:
More info below:

What I confirmed?
```
Aug 25 11:57:33 worker1 systemd[1]: Stopped run-p2811-i2812.service - [systemd-run] /usr/bin/systemctl reboot "--message=Initiated by bootc".
Aug 25 11:57:37 worker1 systemd[1]: systemd-soft-reboot.service: Deactivated successfully.
Aug 25 11:57:37 worker1 systemd[1]: Finished systemd-soft-reboot.service - Reboot System Userspace.
Aug 25 11:57:37 worker1 systemd[1]: Reached target soft-reboot.target - Reboot System Userspace.
Aug 25 11:57:37 worker1 systemd[1]: Stopped target soft-reboot.target - Reboot System Userspace.
```
Also
```
Boot ID BEFORE: 2ee4c375-c4eb-4ffc-998d-112ad41e7f1f
Boot ID AFTER:  2ee4c375-c4eb-4ffc-998d-112ad41e7f1f
```
Also
bn.spec.rebootPolicy: AllowSoftReboot was set on the BootcNode by the controller

Full reboot verification(different)

```
Aug 25 12:36:54 worker2 systemd[1]: Reload requested from client PID 1044 ('systemctl') (unit cloud-init-main.service)...
Aug 25 12:37:03 worker2 systemd[1]: Reload requested from client PID 1239 ('systemctl') (unit session-4.scope)...
Aug 25 12:38:11 worker2 systemd[1]: Shutting down.
Aug 25 12:38:20 worker2 kernel: ISO 9660 Extensions: Microsoft Joliet Level 3
```
Also
```
Boot ID BEFORE: 6a9276a4-c6d7-4353-bb47-286c14b87df0
Boot ID AFTER:  46decf0b-f5b8-4a6e-90bb-6a4dc55986a8
```
Closes https://github.com/bootc-dev/bootc-operator/issues/117

### DIFF
--- a/internal/bootc/executor.go
+++ b/internal/bootc/executor.go
@@ -20,6 +20,7 @@ type Executor interface {
 	Status(ctx context.Context) ([]byte, error)
 	Stage(ctx context.Context, image string) error
 	Reboot(ctx context.Context) error
+	Apply(ctx context.Context, softReboot bool) error
 }
 
 // HostExecutor runs bootc commands on the host via nsenter.
@@ -149,6 +150,23 @@ func (e *HostExecutor) Reboot(ctx context.Context) error {
 	out, err := cmd.CombinedOutput()
 	if err != nil {
 		return fmt.Errorf("running systemctl reboot: %s: %w", out, err)
+	}
+	return nil
+}
+
+func (e *HostExecutor) Apply(ctx context.Context, softReboot bool) error {
+	log := logf.FromContext(ctx)
+
+	args := []string{"bootc", "upgrade", "--from-downloaded", "--apply"}
+	if softReboot {
+		args = append(args, "--soft-reboot=auto")
+	}
+
+	cmd := e.nsenterCmd(ctx, args...)
+	log.Info("Executing", "cmd", strings.Join(cmd.Args, " "))
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("running bootc apply: %s: %w", out, err)
 	}
 	return nil
 }
--- a/internal/bootc/executor.go
+++ b/internal/bootc/executor.go
@@ -20,7 +20,29 @@ type Executor interface {
 	Status(ctx context.Context) ([]byte, error)
 	Stage(ctx context.Context, image string) error
 	Reboot(ctx context.Context) error
-	Apply(ctx context.Context, softReboot bool) error
+	ApplyUpdate(ctx context.Context, softReboot bool) error
+}
+
+// Centralized bootc command builders.
+
+func bootcStatusArgs() []string {
+	return []string{"bootc", "status", "--json", "--format-version", "1"}
+}
+
+func bootcSwitchArgs(image string) []string {
+	return []string{"bootc", "switch", image}
+}
+
+func bootcApplyUpdateArgs(softReboot bool) []string {
+	args := []string{"bootc", "switch", "--from-downloaded", "--apply"}
+	if softReboot {
+		args = append(args, "--soft-reboot=auto")
+	}
+	return args
+}
+
+func systemctlRebootArgs() []string {
+	return []string{"systemctl", "reboot"}
 }
 
 // HostExecutor runs bootc commands on the host via nsenter.
@@ -42,7 +64,7 @@ func (e *HostExecutor) nsenterCmd(ctx context.Context, args ...string) *exec.Cmd
 }
 
 func (e *HostExecutor) Status(ctx context.Context) ([]byte, error) {
-	cmd := e.nsenterCmd(ctx, "bootc", "status", "--json", "--format-version", "1")
+	cmd := e.nsenterCmd(ctx, bootcStatusArgs()...)
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, fmt.Errorf("running bootc status: %w", err)
@@ -68,13 +90,13 @@ func (e *HostExecutor) Stage(ctx context.Context, image string) error {
 	// Ideally we'd use systemd-run's `--pipe` here, which would avoid
 	// having to fetch the unit journal down below, but SELinux blocks it
 	// (dbus-broker can't access container-labeled fds).
-	cmd := e.nsenterCmd(ctx,
-		"systemd-run", "--wait", "--collect",
-		"--unit", stageUnitName,
-		// TODO: use --download-only once available
-		// (https://github.com/bootc-dev/bootc/issues/2137)
-		"bootc", "switch", image,
+	// TODO: use --download-only once available
+	// (https://github.com/bootc-dev/bootc/issues/2137)
+	stageArgs := append(
+		[]string{"systemd-run", "--wait", "--collect", "--unit", stageUnitName},
+		bootcSwitchArgs(image)...,
 	)
+	cmd := e.nsenterCmd(ctx, stageArgs...)
 	cmd.Cancel = func() error {
 		e.stopStageUnit()
 		return nil
@@ -145,7 +167,7 @@ func (e *HostExecutor) copyJournalUnitLogs(log logr.Logger, unit string, cursor 
 func (e *HostExecutor) Reboot(ctx context.Context) error {
 	log := logf.FromContext(ctx)
 
-	cmd := e.nsenterCmd(ctx, "systemctl", "reboot")
+	cmd := e.nsenterCmd(ctx, systemctlRebootArgs()...)
 	log.Info("Executing", "cmd", strings.Join(cmd.Args, " "))
 	out, err := cmd.CombinedOutput()
 	if err != nil {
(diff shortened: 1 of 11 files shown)


HarshwardhanPatil07:
cc @alicefr @Johan-Liebert1 @ptalgulk01

internal/bootc/executor.go
```diff
@@ -20,6 +20,7 @@ type Executor interface {
 	Status(ctx context.Context) ([]byte, error)
 	Stage(ctx context.Context, image string) error
 	Reboot(ctx context.Context) error
+	Apply(ctx context.Context, softReboot bool) error
```
Johan-Liebert1: Should be called `ApplyUpdate`, just Apply is a bit ambiguous

internal/bootc/executor.go
```diff
@@ -152,3 +153,20 @@ func (e *HostExecutor) Reboot(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (e *HostExecutor) Apply(ctx context.Context, softReboot bool) error {
+	log := logf.FromContext(ctx)
+
+	args := []string{"bootc", "upgrade", "--from-downloaded", "--apply"}
+	if softReboot {
+		args = append(args, "--soft-reboot=auto")
```
Johan-Liebert1: We should probably have `soft-reboot=required` if it's true? I don't think this is incorrect as this would just perform a normal reboot is soft reboot isn't available

internal/bootc/executor.go
```diff
@@ -152,3 +153,20 @@ func (e *HostExecutor) Reboot(ctx context.Context) error {
 	}
 	return nil
 }
+
+func (e *HostExecutor) Apply(ctx context.Context, softReboot bool) error {
+	log := logf.FromContext(ctx)
+
+	args := []string{"bootc", "upgrade", "--from-downloaded", "--apply"}
```
Johan-Liebert1: I really think we should have a centralized function/constants that have all the list of bootc commands and we just call that function to get the command we need. Not a huge fan of the current scattered approach we have

internal/daemon/reconciler.go
```diff
@@ -137,11 +136,17 @@ func (r *BootcNodeReconciler) Reconcile(
 
 	// Reboot after the status patch so the Rebooting condition is persisted before the node goes down.
 	if res.needsReboot {
-		log.Info("Starting reboot")
-		if err := r.Executor.Reboot(ctx); err != nil {
-			return ctrl.Result{}, fmt.Errorf("reboot: %w", err)
+		if res.softReboot {
+			log.Info("Applying update with soft reboot")
+			if err := r.Executor.Apply(ctx, true); err != nil {
```
Johan-Liebert1: Does this imply that soft reboot can only be performed with `--download-only` option? If yes, then that's not true for boot.c `bootc update/switch` both support `--soft-reboot` option

internal/bootc/executor.go
HarshwardhanPatil07: From the bootc source (cli.rs):
- Required — "Require a soft reboot; fail if not possible"
- Auto — "Automatically use soft reboot if possible, otherwise use regular reboot"
If we want a RequireSoftReboot policy in the future that fails instead of falling back, we would add a new RebootPolicy enum value and use --soft-reboot=required for that.

internal/daemon/reconciler.go
HarshwardhanPatil07: No, it does not require --download-only. The --from-downloaded flag here just tells bootc to use the already-staged deployment without fetching from the registry again.

But it is not supported by operator It bypasses the drain. It does a full reboot in that case. cc @alicefr Can you please confirm this?

HarshwardhanPatil07: it combines download + stage + reboot in one step in that case

internal/bootc/executor.go
HarshwardhanPatil07: cc @alicefr should we support `soft-reboot=required`?

internal/daemon/reconciler.go
HarshwardhanPatil07: @jlebon did had the same thing in mind i guess. ref: https://github.com/bootc-dev/bootc-operator/issues/117

internal/bootc/executor.go
HarshwardhanPatil07: Yes, Thank you!

internal/bootc/executor.go
HarshwardhanPatil07: Yes, Thank you!

internal/bootc/executor.go
cheesesashimi: Can `bootc` identify whether a soft-reboot is possible after staging an update? Could the controller potentially identify this condition before signalling to the daemon that it should stage the image? The reason I ask is because we should surface the error message via an API object whenever a soft-reboot is not possible and `--soft-reboot=required` is set. And ideally, it would be great to uncover this before staging the OS image.

HarshwardhanPatil07: @cheesesashimi 

> Can bootc identify whether a soft reboot is possible after staging?

Yes

HarshwardhanPatil07: > Could the controller potentially identify this condition before signalling to the daemon that it should stage the image?

Generally no. Bootc needs the staged deployment to compare it with the running system. It cannot reliably know before downloading and preparing the target image.

HarshwardhanPatil07: Yes it will be great, bootc determines this after staging and exposes it as `status.staged.softRebootCapable`, which the daemon already propagates to `BootcNode.status`. `RequireSoftReboot` policy, MCO or the bootc-operator controller could use that status as a gate before selecting and draining the node, surface `SoftRebootUnavailable` through an API condition, and still use `--soft-reboot=required` as final enforcement.

Johan-Liebert1: > Generally no. Bootc needs the staged deployment to compare it with the running system. It cannot reliably know before downloading and preparing the target image.

usually for upgrades, and upgrades only, I think we compute at image build time whether this particular image upgrade would be soft rebootable. It's just a matter of comparing the current and the upgrade image being built. That being said, I don't think there's a good requirement for this

internal/bootc/executor.go
alicefr: I would wait until we have a new build for bink and not adding this additional code

internal/daemon/reconciler.go
alicefr: @HarshwardhanPatil07 @Johan-Liebert1 why cannot we let the operator reboots the node via `systemctl reboot`? As far as I understand, systemd is able to soft reboot, or doesn't bootc switch prepare for the next deployment? I'm not sure how bootc handle this case for soft reboot

test/e2e/bootcnode_test.go
```diff
@@ -582,6 +584,151 @@ func TestNonExistingImage(t *testing.T) {
...
+			}
+		}
+		return ""
+	}).WithTimeout(2*time.Minute).WithPolling(2*time.Second).ShouldNot(BeEmpty(),
+		"running daemon pod not found on %s", nodeName)
+
+	kubeconfigPath := os.Getenv("KUBECONFIG")
+	cmd := exec.CommandContext(ctx, "kubectl", "--kubeconfig", kubeconfigPath,
+		"-n", "bootc-operator", "exec", podName, "--",
+		"cat", "/proc/sys/kernel/random/boot_id")
```
alicefr, line 725: can we read this file using nsenter? The procs filesystem, in this case, is from the container not the host

internal/daemon/reconciler.go
HarshwardhanPatil07: @alicefr `systemctl reboot` from the operator will skip the bootc-specific preparation. 

on composefs, before calling the systemctl soft-reboot it prepares the target root under `/run/nextroot`

On OSTree, bootc calls deployment_set_soft_reboot() before

HarshwardhanPatil07: bootc must first prepare which staged deployment systemd should enter and verify that the transition is safe.

HarshwardhanPatil07: i burnt some tokens for research to compare three repos

HarshwardhanPatil07:
Addressed

internal/daemon/reconciler.go
Johan-Liebert1: > @HarshwardhanPatil07 @Johan-Liebert1 why cannot we let the operator reboots the node via systemctl reboot? As far as I understand, systemd is able to soft reboot, or doesn't bootc switch prepare for the next deployment? I'm not sure how bootc handle this case for soft reboot

bootc, does, in the end call `systemctl soft-reboot`, but we need to prepare the `nextroot` that we want to soft reboot into which consists of mounting the EROFS, preforming the /etc merge, and setting up composefs/ostree states depending on the backend

HarshwardhanPatil07:
FYI if you see ci failing, its because changed the cmd to bootc switch instead of bootc upgrade 

bootc-dev/bootc:2342 pr will land shortly in rpm

PR #152, opened 2026-08-27
e2e: add controller recovery test

ptalgulk01:
Summary

- Add TestControllerRecovery e2e test that verifies the controller recovers after being killed and completes a pending rollout
- The test scales the controller deployment to zero, patches the pool to a new image, then restores the controller and asserts the rollout completes

Context

Part of the e2e test planning in #69 (scenario 5: kill controller/daemon during rollout).

Manual testing also revealed that scenarios 4 (faulty node after reboot) and 8 (cross-distro bad image) both fail — the controller has no reboot timeout and doesn't correlate Node.Ready status, so nodes stuck after reboot are never marked degraded. Those need feature work before tests can be written.

Test plan

- [x] make e2e V=1 RUN=TestControllerRecovery passes locally (159s)

### DIFF
--- a/test/e2e/bootcnode_test.go
+++ b/test/e2e/bootcnode_test.go
@@ -12,6 +12,7 @@ import (
 
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


test/e2e/bootcnode_test.go
```diff
@@ -582,6 +583,134 @@ func TestNonExistingImage(t *testing.T) {
...
+		HaveField("Booted.ImageDigest", Equal(env.NodeImageDigest())),
+		"node should not update while controller is down",
+	)
+
+	// Phase 4: Scale controller back to one.
+	g.Expect(env.Client.Get(ctx, deployKey, controllerDeploy)).To(Succeed())
+	modified = controllerDeploy.DeepCopy()
+	one := int32(1)
+	modified.Spec.Replicas = &one
+	g.Expect(env.Client.Patch(ctx, modified, client.MergeFrom(controllerDeploy))).To(Succeed())
```
HarshwardhanPatil07: Test never interrupts an active rollout

The image changes while the controller is already stopped, so restart simply begins a new rollout.

HarshwardhanPatil07 (changes requested):

HarshwardhanPatil07 (commented):
You will need to sign DCO

test/e2e/bootcnode_test.go
```diff
@@ -582,6 +583,170 @@ func TestNonExistingImage(t *testing.T) {
...
+			HaveField("Reason", Or(
+				Equal(bootcv1alpha1.NodeReasonStaging),
+				Equal(bootcv1alpha1.NodeReasonStaged),
+			)),
+		))),
+	), "expected node to be staging (rollout in progress, pre-reboot) before killing controller")
+
+	t.Logf("Rollout in progress; interrupting the controller")
+
+	// Phase 4: Scale the controller to zero, interrupting the active rollout.
```
alicefr, line 658: I'm wondering if we should pause the pool at this stage. I'm afraid that if the pulling is fast enough, the node might go in rebooting before scale down the controller, and it is a possible source for race conditions. Would it make sense to pause it when we detect the staging, scale down the operator and unpause it?

test/e2e/bootcnode_test.go
```diff
@@ -582,6 +583,170 @@ func TestNonExistingImage(t *testing.T) {
...
+	g.Expect(env.Client.Patch(ctx, down, client.MergeFrom(controllerDeploy))).To(Succeed())
+
+	g.Eventually(func() (int32, error) {
+		var d appsv1.Deployment
+		err := env.Client.Get(ctx, deployKey, &d)
+		return d.Status.AvailableReplicas, err
+	}).WithTimeout(1*time.Minute).Should(BeZero(),
+		"expected controller to scale to zero")
+
+	t.Logf("Controller scaled to zero mid-rollout")
```
alicefr, line 692: Can you create an helper function for scaling up and down the controller? It could be reused by other tests and it makes the entire test more readable

test/e2e/bootcnode_test.go
```diff
@@ -582,6 +583,170 @@ func TestNonExistingImage(t *testing.T) {
...
+	g.Expect(env.Client.Patch(ctx, up, client.MergeFrom(controllerDeploy))).To(Succeed())
+
+	g.Eventually(func() (int32, error) {
+		var d appsv1.Deployment
+		err := env.Client.Get(ctx, deployKey, &d)
+		return d.Status.AvailableReplicas, err
+	}).WithTimeout(2*time.Minute).Should(Equal(int32(1)),
+		"expected controller to scale back to one")
+
+	t.Logf("Controller restored; waiting for the interrupted rollout to finish")
```
alicefr, line 724: This also could be replaced by the helper function

ptalgulk01: Done

test/e2e/bootcnode_test.go
ptalgulk01: Done

test/e2e/bootcnode_test.go
ptalgulk01: yes it does, I have updated, thankyou for pointing out.

alicefr (approved):

HarshwardhanPatil07 (approved):
lgtm

PR #155, opened 2026-09-02
Emit Kubernetes Events

### LINKED ISSUE #101: Emit Kubernetes Events
Currently, we don't create any kubernetes events, but we could fire an event:
For the BootcNodePool
* when rollout starts and completes
* If the pool goes in degraded state, like there is a node conflict or there in an invalid spec
* There is a new version of the  tagged image and this triggers a rollout
* Rollout is halted
For the BoocNode when a node:
* is in staging
* is in staged
* is rebooting

The events should always be fired by the controller and not by the daemon, in this way we keep the daemon simple and we can limit the RBAC to the event to the controller only.

HarshwardhanPatil07:
Pool Events:

- `ImageUpdateAvailable`
- `RolloutStarted`
- `RolloutCompleted`
- Warning Events when the pool enters or changes a degraded state

Node Events:

- `Staging`
- `Staged`
- `Rebooting`
- `DrainFailed`
- `DrainTakingTooLong`

Node Events reference the BootcNode as the primary object and the owning BootcNodePool as the related object.

## Testing

Completed: Manual Event verification on a bink cluster

```shell
make buildimg
make deploy-bink
make e2e V=1 RUN='TestUpdateReboot\|TestTagResolution'
```

watch events in second terminal: `kubectl get events.events.k8s.io -A --watch`

closes: #101

### DIFF
--- a/api/v1alpha1/constants.go
+++ b/api/v1alpha1/constants.go
@@ -25,4 +25,9 @@ const (
 	// the K8s Node was already cordoned before the controller cordoned
 	// it for a reboot. Used to restore prior cordon state after update.
 	AnnotationWasCordoned = "bootc.dev/was-cordoned"
+
+	// AnnotationLastObservedState records the last BootcNode Idle condition
+	// state for which the controller considered emitting an event. It is
+	// observability bookkeeping only and is never used to drive reconciliation.
+	AnnotationLastObservedState = "bootc.dev/last-observed-state"
 )
--- a/internal/controller/events.go
+++ b/internal/controller/events.go
@@ -4,6 +4,8 @@ package controller
 
 import (
 	"fmt"
+	"strings"
+	"unicode/utf8"
 
 	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
@@ -20,6 +22,10 @@ const (
 
 	eventActionResolveImage = "ResolveImage"
 	eventActionRollout      = "Rollout"
+	eventActionPoolDegraded = "PoolDegraded"
+
+	eventNoteLimit  = 1024
+	eventNoteSuffix = "..."
 )
 
 // EventNote renders the human-readable message of a Kubernetes Event. Each
@@ -73,6 +79,16 @@ type poolRolloutCompletedNote struct {
 
 func (n poolRolloutCompletedNote) Note() string {
 	return fmt.Sprintf("Rollout completed at digest %s", shortDigest(n.TargetDigest))
+}
+
+// poolDegradedNote wraps a pool's Degraded condition message, which is not
+// length-bounded, so it caps itself.
+type poolDegradedNote struct {
+	Message string
+}
+
+func (n poolDegradedNote) Note() string {
+	return capNote(n.Message)
 }
 
 func (r *BootcNodePoolReconciler) recordPoolEvents(
@@ -133,6 +149,19 @@ func (r *BootcNodePoolReconciler) recordPoolEvents(
 			poolRolloutCompletedNote{TargetDigest: pool.Status.TargetDigest},
 		)
 	}
+
+	oldDegraded := apimeta.FindStatusCondition(previous.Conditions, bootcv1alpha1.PoolDegraded)
+	newDegraded := apimeta.FindStatusCondition(pool.Status.Conditions, bootcv1alpha1.PoolDegraded)
+	if degradedConditionChanged(oldDegraded, newDegraded) {
+		r.recordEvent(
+			pool,
+			nil,
+			corev1.EventTypeWarning,
+			newDegraded.Reason,
+			eventActionPoolDegraded,
+			poolDegradedNote{Message: newDegraded.Message},
+		)
+	}
 }
 
 func conditionEnteredReason(
@@ -146,10 +175,37 @@ func conditionEnteredReason(
 	return previous == nil || previous.Status != status || previous.Reason != reason
 }
 
+func degradedConditionChanged(previous, current *metav1.Condition) bool {
+	if current == nil || current.Status != metav1.ConditionTrue {
+		return false
+	}
+	return previous == nil ||
+		previous.Status != metav1.ConditionTrue ||
+		previous.Reason != current.Reason ||
+		previous.Message != current.Message
+}
+
 func (r *BootcNodePoolReconciler) recordEvent(
 	regarding, related runtime.Object,
 	eventType, reason, action string,
 	note EventNote,
 ) {
 	r.Recorder.Eventf(regarding, related, eventType, reason, action, "%s", note.Note())
+}
+
+// capNote keeps a note within the events.k8s.io/v1 1 KiB limit and never splits
+// a UTF-8 sequence. It is only needed for notes built from unbounded free text
+// (condition messages, error strings); notes assembled from bounded fields fit
+// by construction.
+func capNote(note string) string {
+	note = strings.ToValidUTF8(note, "�")
+	if len(note) <= eventNoteLimit {
+		return note
+	}
+
+	limit := eventNoteLimit - len(eventNoteSuffix)
+	for limit > 0 && !utf8.RuneStart(note[limit]) {
+		limit--
+	}
+	return note[:limit] + eventNoteSuffix
 }
--- a/internal/controller/events.go
+++ b/internal/controller/events.go
@@ -1,0 +1,155 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+
+	bootcv1alpha1 "github.com/bootc-dev/bootc-operator/api/v1alpha1"
+)
+
+const (
+	eventReasonImageUpdateAvailable = "ImageUpdateAvailable"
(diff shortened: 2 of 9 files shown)


HarshwardhanPatil07:
PTAL @Johan-Liebert1 @ptalgulk01 @alicefr

test/e2e/bootcnode_test.go
```diff
@@ -594,7 +660,31 @@ func fetchPoolStatus(
...
+
+		events := make([]eventsv1.Event, 0, len(eventList.Items))
+		for _, event := range eventList.Items {
+			if event.Regarding.Kind == kind &&
+				event.Regarding.Name == name &&
+				event.Regarding.UID == uid {
+				events = append(events, event)
+			}
+		}
+		return events, nil
```
ptalgulk01: we can have this as helper func something like 
```
func FilterEventsByObject(items []eventsv1.Event, kind, name string, uid types.UID) []eventsv1.Event {
      out := make([]eventsv1.Event, 0, len(items))
      for _, e := range items {
              if e.Regarding.Kind == kind && e.Regarding.Name == name && e.Regarding.UID == uid {
                      out = append(out, e)
              }
      }
      return out
}

```

internal/controller/events.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	observation = fmt.Sprintf("%s:%s", idle.Status, idle.Reason)
+	if idle.Status != metav1.ConditionFalse {
+		return observation, "", ""
+	}
+
+	switch idle.Reason {
+	case bootcv1alpha1.NodeReasonStaging:
+		return observation + ":" + node.Spec.DesiredImage,
```
ptalgulk01: we can have this store in var `observation + ":" + node.Spec.DesiredImage`

HarshwardhanPatil07:
@ptalgulk01 Thanks. added the changes

HarshwardhanPatil07:
rebase

api/v1alpha1/constants.go
```diff
@@ -25,4 +25,9 @@ const (
 	// the K8s Node was already cordoned before the controller cordoned
 	// it for a reboot. Used to restore prior cordon state after update.
 	AnnotationWasCordoned = "bootc.dev/was-cordoned"
+
+	// AnnotationLastObservedState records the last BootcNode Idle condition
+	// state for which the controller considered emitting an event. It is
+	// observability bookkeeping only and is never used to drive reconciliation.
+	AnnotationLastObservedState = "bootc.dev/last-observed-state"
```
alicefr: I don't quite understand for what this annotation is used, can you please explain

internal/controller/events.go
```diff
@@ -0,0 +1,323 @@
...
+
+	drainStallThreshold = 5 * time.Minute
+	eventNoteLimit      = 1024
+	eventNoteSuffix     = "..."
+)
+
+// updatePoolStatus writes a changed pool status and then records events for
+// meaningful transitions. Events are supplemental: a failed status write
+// returns before any event is emitted.
+func (r *BootcNodePoolReconciler) updatePoolStatus(
```
alicefr: Does this need to be move in the event file?

internal/controller/events.go
```diff
@@ -0,0 +1,323 @@
...
+		previous.TargetDigest != pool.Status.TargetDigest &&
+		pool.Status.TargetDigest != ""
+	if conditionEnteredReason(
+		oldUpToDate,
+		newUpToDate,
+		metav1.ConditionFalse,
+		bootcv1alpha1.PoolRolloutInProgress,
+	) || (targetChanged && newUpToDate != nil &&
+		newUpToDate.Status == metav1.ConditionFalse &&
+		newUpToDate.Reason == bootcv1alpha1.PoolRolloutInProgress) {
```
alicefr: can you summarize this in a variable with a meaningful name like `poolRolloutInProgress`

internal/controller/events.go
```diff
@@ -0,0 +1,323 @@
...
+			eventReasonRolloutStarted,
+			eventActionRollout,
+			"Rollout started toward digest %s",
+			pool.Status.TargetDigest,
+		)
+	}
+	if oldUpToDate != nil &&
+		oldUpToDate.Status != metav1.ConditionTrue &&
+		newUpToDate != nil &&
+		newUpToDate.Status == metav1.ConditionTrue {
```
alicefr: same here, like `rolloutCompleted`

internal/controller/events.go
```diff
@@ -0,0 +1,323 @@
...
+			// hide the transition. A retry may aggregate the same event into an
+			// EventSeries, which is preferable to losing it.
+			log.Error(err, "Failed to persist last observed node state", "node", node.Name)
+			continue
+		}
+		*node = *modified
+	}
+}
+
+func nodeEvent(node *bootcv1alpha1.BootcNode) (observation, reason, note string) {
```
alicefr: Maybe better to define a struct here to reduce the number of returned arguments

internal/controller/events.go
```diff
@@ -0,0 +1,323 @@
...
+	eventType, reason, action, note string,
+	args ...any,
+) {
+	note = truncateEventNote(fmt.Sprintf(note, args...))
+	r.Recorder.Eventf(regarding, related, eventType, reason, action, "%s", note)
+}
+
+// truncateEventNote keeps notes within the events.k8s.io/v1 1 KiB limit and
+// never splits a UTF-8 sequence.
+func truncateEventNote(note string) string {
```
alicefr: Is there a better way to guarantee that the note doesn't bypass the 1KiB. You might just cut the output. Don't we controll how the note looks like

alicefr:
@HarshwardhanPatil07 can we split the commit 161e5e7d37006707dee96e3cb3f9fb12ff48c384 into mulitple. At least one per event type, like one for the draining stalls, one for the node events and one for the degraded nodes.

internal/controller/events.go
HarshwardhanPatil07: bootcnodepool_controller.go moved here, controller logic

internal/controller/events.go
HarshwardhanPatil07: added note

internal/controller/events.go
```diff
@@ -0,0 +1,125 @@
...
+	args ...any,
+) {
+	note = truncateEventNote(fmt.Sprintf(note, args...))
+	r.Recorder.Eventf(regarding, related, eventType, reason, action, "%s", note)
+}
+
+// truncateEventNote keeps notes within the events.k8s.io/v1 1 KiB limit and
+// never splits a UTF-8 sequence. Event templates are fixed, but interpolated
+// image references, condition messages, and errors are not length-bounded.
+func truncateEventNote(note string) string {
```
alicefr: I'm really not a big fan of truncate the string. I think instead we should use an interface and struct with well defined fields and functions that can shorten the string. Something like:
```golang
 // Event note types — each knows its own max rendered length.
  type PoolImageUpdateNote struct {
      ImageRef       string // max ~255 (OCI spec)
      NewDigest      string // fixed 71 (sha256:64hex)
      PreviousDigest string // fixed 71
  }

  func (n PoolImageUpdateNote) Note() string {
      return fmt.Sprintf("Image tag %s resolved to new digest %s (previously %s)",
          n.ImageRef, shortDigest(n.NewDigest), shortDigest(n.PreviousDigest))
  }

  type PoolRolloutStartedNote struct {
      TargetDigest string
  }

  func (n PoolRolloutStartedNote) Note() string {
      return fmt.Sprintf("Rollout started toward digest %s", shortDigest(n.TargetDigest))
  }

  // Interface all event notes implement.
  type EventNote interface {
      Note() string
  }

  func shortDigest(d string) string {
      if len(d) > 19 { // "sha256:" + 12 hex
          return d[:19]
      }
      return d
  }
``` 
and then the `recordEventf` becomes:
```golang
func (r *BootcNodePoolReconciler) recordEvent(
      regarding, related runtime.Object,
      eventType, reason, action string,
      note EventNote,) 
{
      r.Recorder.Eventf(regarding, related, eventType, reason, action, "%s", note.Note())
}
```

internal/controller/events.go
```diff
@@ -124,6 +128,96 @@ func degradedConditionChanged(previous, current *metav1.Condition) bool {
...
+		return nodeEventData{}
+	}
+
+	observation := fmt.Sprintf("%s:%s", idle.Status, idle.Reason)
+	if idle.Status != metav1.ConditionFalse {
+		return nodeEventData{observation: observation}
+	}
+
+	observationWithImage := observation + ":" + node.Spec.DesiredImage
+	switch idle.Reason {
```
alicefr, line 323: Should we also add when the node becomes idle again?

HarshwardhanPatil07: added a nodeIdleNote "Node is up to date" event emitted when a node returns to idle

HarshwardhanPatil07:
PTAL @alicefr

internal/controller/events.go
```diff
@@ -185,6 +221,104 @@ func degradedConditionChanged(previous, current *metav1.Condition) bool {
...
+				note,
+			)
+		}
+
+		modified := node.DeepCopy()
+		if modified.Annotations == nil {
+			modified.Annotations = map[string]string{}
+		}
+		modified.Annotations[bootcv1alpha1.AnnotationLastObservedState] = observation
+		if err := r.Patch(ctx, modified, client.MergeFrom(node)); err != nil {
```
alicefr, line 258: Don't we need to check if the observation has changed before patching?

alicefr (approved):

internal/controller/events.go
HarshwardhanPatil07: it's already present on line 264

HarshwardhanPatil07:
Thank you @alicefr for the review!

PR #159, opened 2026-09-03
controller: mark nodes stuck rebooting as degraded

### LINKED ISSUE #158: Detect nodes that never recover after reboot (reboot timeout → degraded)
## Summary

The operator has no mechanism to detect a node that fails to come back after a
reboot. Once a node enters `Rebooting`, if it never returns (bad image, hardware
fault, node shutdown), it stays in that state indefinitely — the BootcNode is
never marked `Degraded` and the reboot slot is held forever.

This blocks two scenarios from #69:

- **#69 item 4** — "Shutdown (manually) a node after the reboot to simulate a
  faulty node, after the upgrade. **It should be reported as degraded**."
  → Today it is **not** reported as degraded.
- **#69 item 8** — "Switch to a completely different image e.g. fedora to centos
  stream." → Node reboots into a kubelet-less image and gets stuck in
  `Rebooting` forever. (Expected behavior for this case is undefined and should
  be decided as part of this work.)

## Reproduction (verified on a bink cluster)

**Faulty node after upgrade (#69 item 4):** node updates successfully
(`updatedCount=1`), then the VM is stopped. Observed, stable for 6+ min:

| Signal | Value |
|--------|-------|
| Node `Ready` | Unknown |
| BootcNode `Degraded` | False |
| Pool `degradedCount` | 0 |

**Cross-distro bad image, no kubelet (#69 item 8):** node reboots into the bad
image and never returns. Observed, stable for 5+ min:

| Signal | Value |
|--------|-------|
| Node `Ready` | Unknown |
| BootcNode `Idle` | False (`Rebooting`) |
| Pool `degradedCount` | 0 |

Controller log confirms the slot is wedged:
`Rollout state ... rebooting: 1, degraded: 0, occupiedSlots: 1, availableSlots: 0, candidates: []`

## Root cause

- `classifyNode()` (`internal/controller/rollout.go:614`) classifies a node as `rebooting` while `Idle=False/Reason=Rebooting` (`:658`). There is no time  bound on this state.
- `findUnhealthySlots()` (`rollout.go:460`) only scans `degraded` and `upToDate` nodes. A node stuck in `rebooting` is in neither list, so it is never counted as unhealthy.
- The halt logic requires `>= unhealthySlotHaltThreshold` (== 2, `rollout.go:34,96`) unhealthy nodes, so a single stuck node never halts the rollout.
- The daemon sets `Rebooting` then calls `systemctl reboot` (`internal/bootc/executor.go:144`); if the node never returns it cannot report anything because it is down. Detection must therefore be **controller-side**.

## Proposed change

1. **Reboot timeout (controller-side).** Track time since a node entered `Rebooting` (the `Idle=False/Rebooting` condition's `LastTransitionTime`). If  it exceeds a timeout, transition the BootcNode to `Degraded` with a new reason `RebootTimeout`.
2. **API.** Add `RebootTimeoutSeconds` to `DisruptionSpec`/`RolloutSpec` (default ~10 min) and a `NodeReasonRebootTimeout` constant.
3. **Feed existing halt logic.** Timed-out nodes become `degraded`, which flows  into `findUnhealthySlots()` / `PoolRolloutHalted` (interacts with #99).
4. **Decide #69 item 8 semantics.** Define expected behavior for a node that boots an image without kubelet (reject upfront? same reboot-timeout →degraded?).

## Related

- #69 (E2e test planning — items 4 and 8)
- #99 (configurable rollout halt threshold)

ptalgulk01:
## What

Adds `spec.rollout.rebootTimeoutSeconds` to BootcNodePool. If a node stays in
the `Rebooting` state longer than this (e.g. it failed to boot the new image or
never rejoined the cluster), the controller reports it as degraded at the pool
level (`degradedCount` + `Degraded/NodeDegraded`). If 2+ nodes time out, the
rollout halts (`Degraded/RolloutHalted`).

The field is optional; when unset the controller waits indefinitely, so existing
behavior is unchanged.

## Why

Today a node that reboots and never comes back sits in `Rebooting` forever. The
daemon can't report it (the node is down), so nothing surfaces the failure and
the reboot slot is held indefinitely. This covers issue #69 item 4 and the
"cross-distro bad image" scenario.
Fixes: https://github.com/bootc-dev/bootc-operator/issues/158 
## How it works

- `driveRollout` reclassifies timed-out `Rebooting` nodes into the degraded
  bucket and requeues the pool so the timeout fires without needing an event.
- `findUnhealthySlots` counts timed-out nodes toward the halt threshold — a node
  that failed to boot the target is a strong signal the image is bad, which is
  what the halt is meant to catch.

## Design note

The controller reports at the **pool level only**. It has read-only access to
`bootcnodes/status`, and the daemon owns that condition (it auto-clears Degraded
on recovery). Marking the BootcNode's own condition would need an RBAC change and
a second writer for daemon-owned status. Happy to go that route instead if
maintainers prefer — flagging it here for input.

## Testing

Unit tests for the timeout decision and slot classification, plus envtests for
the pool-degraded and rollout-halt paths.
Assisted-by: AI


internal/controller/bootcnodepool_controller.go
```diff
@@ -322,6 +322,13 @@ func (r *BootcNodePoolReconciler) Reconcile(
 	// but counts catch up on the next successful reconcile.
 	syncPoolStatus(&pool, rs)
 
+	// Carry a time-based requeue (e.g. reboot timeout) unless something
+	// sooner is already scheduled.
+	if rs.requeueAfter > 0 &&
+		(resolveResult.RequeueAfter == 0 || rs.requeueAfter < resolveResult.RequeueAfter) {
+		resolveResult.RequeueAfter = rs.requeueAfter
+	}
```
alicefr: why is this necessary?

alicefr (approved):

alicefr:
It looks great, many thanks for this work! Nice coverage also with the env tests

PR #164, opened 2026-09-11
pull secret propagation and registry authentication - milestone 5b

### LINKED ISSUE #135: Pull secret propagation (Milestone-5b)
Implement pull secret propagation so that nodes can pull from authenticated registries.

alicefr:
Implementation of the milestone 5b. It enables the authentication for the container registry and the pull secret propagation

Fixes: #135


alicefr:
@HarshwardhanPatil07 PTAL

internal/controller/bootcnodepool_controller.go
```diff
@@ -161,6 +163,39 @@ func (r *BootcNodePoolReconciler) mapNodeToPoolRequests(
...
+// mapSecretToPoolRequests maps a Secret event to the BootcNodePool(s) that
+// should be reconciled. It enqueues pools whose pullSecretRef matches the
+// changed Secret's name and namespace.
+func (r *BootcNodePoolReconciler) mapSecretToPoolRequests(
+	ctx context.Context,
+	obj client.Object,
+) []reconcile.Request {
+	secret, ok := obj.(*corev1.Secret)
+	if !ok {
+		return nil
```
Johan-Liebert1, line 177: Swallowing the reflection error here. This this deserves an error log

internal/controller/bootcnodepool_controller.go
```diff
@@ -219,6 +254,18 @@ func nodeUnschedulableChanged(oldNode, newNode *corev1.Node) bool {
 	return oldNode.Spec.Unschedulable != newNode.Spec.Unschedulable
 }
 
+// secretTypePredicate returns a predicate that accepts only Secrets of
+// type kubernetes.io/dockerconfigjson.
+func secretTypePredicate() predicate.Predicate {
+	return predicate.NewPredicateFuncs(func(obj client.Object) bool {
+		secret, ok := obj.(*corev1.Secret)
+		if !ok {
+			return false
```
Johan-Liebert1, line 266: same as above

internal/controller/bootcnodepool_controller.go
```diff
@@ -269,8 +316,12 @@ func (r *BootcNodePoolReconciler) Reconcile(
 	// Status writes to `pool` are also permitted; we Update() back once at
 	// the end.
 
-	// Resolve the target digest from the image ref.
-	resolveResult, err := r.resolveTargetDigest(ctx, &pool)
+	secretData, err := r.fetchPullSecretData(ctx, &pool)
+	if err != nil {
+		log.Error(err, "Failed to fetch pull secret")
+		setPoolDegraded(&pool, bootcv1alpha1.PoolSecretError, err.Error())
```
Johan-Liebert1, line 326: missing `return err` here? The following resolution will fail if secret doesn't exist right?

internal/controller/bootcnodepool_controller.go
```diff
@@ -786,10 +870,11 @@ func (r *BootcNodePoolReconciler) restoreCordonState(
 var poolDegradedPrecedence = map[string]int{
 	bootcv1alpha1.PoolHealthy:            0,
```
Johan-Liebert1, line 905: nit: should this be here? Healthy means it's not degraded

internal/registry/resolver.go
```diff
@@ -18,25 +20,61 @@ type GGCRResolver struct {
...
+
+// dockerConfigJSON mirrors the structure of a Kubernetes
+// kubernetes.io/dockerconfigjson secret's .dockerconfigjson key.
+type dockerConfigJSON struct {
+	Auths map[string]authn.AuthConfig `json:"auths"`
+}
+
+// staticKeychain resolves credentials from a parsed dockerconfigjson.
+type staticKeychain struct {
+	auths map[string]authn.AuthConfig
```
Johan-Liebert1: Why can't we use `dockerConfigJSON` for this? Is it to keep `auths` private?

api/v1alpha1/bootcnodepool_types.go
```diff
@@ -56,6 +56,10 @@ const (
 	// resolve the tag.
 	PoolTagResolutionError string = "TagResolutionError"
 
+	// PoolSecretError means the referenced pull secret is missing or
```
Johan-Liebert1, line 59: Maybe two separate errors would make sense here; one for missing secret and one for unexpected key?

internal/controller/bootcnodepool_controller.go
alicefr: It was for not blocking the rollout in case the image was effectively public, but you are right it is misleading

internal/registry/resolver.go
alicefr: Good point, no reason for the extra type. We can include the Keychain interface onto dockerConfigJSON directly.

api/v1alpha1/bootcnodepool_types.go
alicefr: The message field already include the differentiation. Splitting the reason constant wouldn't change controller behavior since both are non-recoverable without user action. Keeping one reason keeps the API simpler, similar to how TagResolutionError covers multiple failure modes

internal/controller/bootcnodepool_controller.go
alicefr: Probably the name poolDegradedPrecedence is misleading. Healthy is there as the baseline so any actual degraded reason wins over it.

Johan-Liebert1 (approved):

PR #169, opened 2026-09-15
Abstract cluster provider to add EKS

alicefr:
Abstraction of the cluster provider in order to support simultaneously bink and other kubernetes platform like EKS.


test/e2e/e2eutil/provider_eks.go
```diff
@@ -0,0 +1,350 @@
...
+			LaunchTemplateId: &p.ltID,
+			Version:          &p.ltVersion,
+		},
+		NetworkInterfaces: []ec2types.InstanceNetworkInterfaceSpecification{
+			{
+				DeviceIndex:              aws.Int32(0),
+				SubnetId:                 &p.subnetID,
+				AssociatePublicIpAddress: aws.Bool(true),
+			},
+		},
```
HarshwardhanPatil07, line 127: should we mention the security groups?
 _Security groups under Network settings for the instance or Security groups under Network interfaces (Add network interface), but not both._

https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html

test/e2e/e2eutil/provider_eks_test.go
```diff
@@ -0,0 +1,189 @@
...
+		WithObjects(objs...).
+		Build()
+}
+
+func TestEKSAddNode(t *testing.T) {
+	const (
+		instanceID = "i-0123456789abcdef0"
+		nodeName   = "ip-10-0-1-100.ec2.internal"
+		providerID = "aws:///us-east-1a/" + instanceID
+	)
```
HarshwardhanPatil07, line 65: Should we capture the real Instance input here?

test/e2e/e2eutil/provider_eks.go
```diff
@@ -0,0 +1,350 @@
...
+	)
+	if err != nil {
+		return "", "", fmt.Errorf("describing ASG %q: %w", asgName, err)
+	}
+	if len(asgOut.AutoScalingGroups) == 0 {
+		return "", "", fmt.Errorf("ASG %q not found", asgName)
+	}
+
+	asg := asgOut.AutoScalingGroups[0]
```
HarshwardhanPatil07: sounds similar to DescribeAutoScalingGroups above under findASGByTags()

ptalgulk01:
I checked the new TC added is not seen in CI jobs, is it expected?

alicefr:
> I checked the new TC added is not seen in CI jobs, is it expected?

@ptalgulk01 do you mean a CI for EKS, we still don't have a public one. I have tested this in our downstream repository. I know that's not ideal, but we don't have the ability to add public CI at the moment, but it isn't breaking bink testing

test/e2e/e2eutil/provider_bink.go
```diff
@@ -0,0 +1,84 @@
...
+
+// BinkProvider provisions nodes using bink (KVM-based local clusters).
+type BinkProvider struct {
+	clusterName  string
+	targetImgRef string
+	diskImage    string
+}
+
+// NewBinkProvider creates a BinkProvider for the given bink cluster.
+func NewBinkProvider(
```
cheesesashimi: question (non-blocking): Would someone directly call this or the NewEKSProvider functions or is this intended to be handled completely through the Env struct? If the latter case, maybe we could make these structs and constructors private to this package as a way to signify that one should only have to instantiate the Env struct when writing a new test.

test/e2e/e2eutil/provider.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"context"
+	"os"
+	"testing"
+)
+
+// NodeProvider abstracts node provisioning so the same tests can run
+// against different infrastructure (bink, EKS, etc.).
+type NodeProvider interface {
```
cheesesashimi, line 13: thought (non-blocking): Something I like to do whenever I have an interface with multiple implementations is add a compile-time check, e.g.:

```golang
var _ NodeProvider = (*EKSProvider)(nil)
var _ NodeProvider = (*BinkProvider)(nil)
var _ NodeProvider = (*Env)(nil)
```

It's definitely not required, but I've found that it helps get feedback more quickly whenever I'm adding a new interface that could be subject to change.

test/e2e/e2eutil/provider_eks_test.go
```diff
@@ -0,0 +1,201 @@
...
+	scheme := runtime.NewScheme()
+	_ = corev1.AddToScheme(scheme)
+
+	return fake.NewClientBuilder().
+		WithScheme(scheme).
+		WithObjects(objs...).
+		Build()
+}
+
+func TestEKSAddNode(t *testing.T) {
```
cheesesashimi, line 60: praise: I really like that there are tests for this!

test/e2e/bootcnode_test.go
```diff
@@ -819,6 +825,7 @@ func TestNonExistingImage(t *testing.T) {
 // registry shares storage with the unauthenticated one (port 5000),
 // so the update image is already available at both endpoints.
 func TestPullSecretAuth(t *testing.T) {
+	e2eutil.Providers(t, "bink", "eks")
```
cheesesashimi, line 828: question (non-blocking): Would this be ran via two separate CI jobs where one job targets bink and the other job targets EKS? Or would both be covered via the same CI job?

cheesesashimi (commented):
Overall, this looks good! I don't have any blocking feedback, only minor suggestions and a few questions mostly for my own edification.

test/e2e/bootcnode_test.go
alicefr: It will be 2 separate CI. For now, we don't have a plan to add EKS to the public CI, but if we have some budget I'd love to. Right now it will simply be a downstream CI.

alicefr:
Checking if the CI is happy and then I will squash the last commit.

test/e2e/e2eutil/provider_bink.go
alicefr: Done thanks for the suggestion

alicefr:
@cheesesashimi @HarshwardhanPatil07 CI is green and I solved all the comments. Could you please have a final look to this and if you are fine with it approve it?

HarshwardhanPatil07 (approved):
Thank you very much!

alicefr:
@Mergifyio backport release-0.1

PR #171, opened 2026-09-16
Create the first release for bootc-operator 0.1.0

alicefr:
Generally, the release should be trigger by bumping the  [VERSION](https://github.com/bootc-dev/bootc-operator/blob/main/VERSION) file but this has already been created wrongly with the current release version. Additionally, this adds the possibility to trigger the release manually using the GHA.


alicefr:
@cheesesashimi @HarshwardhanPatil07 @ptalgulk01 @Johan-Liebert1 I would like to propose to cut the first release. 
We still have a single remaining  [issue](https://github.com/bootc-dev/bootc-operator/issues?q=is%3Aissue%20state%3Aopen%20milestone%3Av0.1) in the list for milestone 0.1 which is the testing. But I think we have already covered a good portion for the testing. WDYT?

HarshwardhanPatil07 (dismissed):
Yes @alicefr +1 on that. We should also take into consideration what is remaining from testing side and update the issue in today's meeting

.github/workflows/release.yaml
```diff
@@ -3,14 +3,16 @@ name: Release
 on:
   pull_request:
```
Johan-Liebert1, line 4: Do we not want the following here?

```suggestion
  pull_request:
    branches:
      -main
```

alicefr: Yeah good point

Johan-Liebert1 (approved):

alicefr:
I'd like to cut the release so we can start to test the operator upgrade path and use this as base

PR #175, opened 2026-09-17
gha: use client-id instead of deprecated app-id

alicefr:
The release job is failing with see [here](https://github.com/bootc-dev/bootc-operator/actions/runs/35212679823):
```
2026-09-17T10:51:48.5256403Z Error: The 'client-id' (or deprecated 'app-id') input must be set to a non-empty string. If using a secret or variable, ensure it is available in this workflow context.

```


HarshwardhanPatil07 (approved):

cgwalters:
What I think would be neat is to try to get an org-wide shared release workflow, though that's trickier.

PR #176, opened 2026-09-17
Fix release v2

alicefr:
Align with bcvk release flow: use GPG key import for git identity and signed tags, pass app token to checkout instead of manually setting the remote URL.

Requires GPG_PRIVATE_KEY and GPG_PASSPHRASE secrets.


alicefr:
This change proved to be correct, see https://github.com/bootc-dev/bootc-operator/actions/runs/35220957641/job/105200705029. It created a draft release and I just deleted to test the change on main

HarshwardhanPatil07 (approved):

